PR #31, opened 2015-10-05
(WIP) Use Marionette instead of default FirefoxDriver

marco-c:


gauntface:
Just a heads up, I spent some time trying to get this working on sw-testing-helpers and you might find some of it useful.

- Get Marionette: https://github.com/GoogleChrome/sw-testing-helpers/blob/master/project/get-marionette.sh
- useMarionette(true): https://github.com/GoogleChrome/sw-testing-helpers/blob/c6734fed89c67d83081a305fc5363d5bcaf34dc2/src/node/webdriver-browser/firefox.js#L46

marco-c:
@gauntface thanks! I also have another problem now, I'm running the tests with some old versions of Node.js (see #150) and the newer versions of `selenium-webdriver` don't support them.

gauntface:
Maybe you could avoid the selenium-webdriver tests in 0.10?

If you have tests for expected input -> output in 0.10, that should suffice and then run webdriver for other versions as integration tests.

marco-c:
Yes, my plan was to do that once I had to. I guess now's a good time to drop it.

gauntface:
Might improve travis run times if you do it for a single node version, but again maybe only worth addressing when it becomes a problem

PR #53, opened 2015-10-21
Enable running a subset of the automated tests on OS X

marco-c:


marco-c:
We need a way to run browser on OS X in a headless fashion.

marco-c:
> We need a way to run browsers on OS X in a headless fashion.

Actually, we don't.

PR #65, opened 2015-12-22
Resolve/reject promise with response information

### LINKED ISSUE #56: sendNotification should throw a meaningful error when the status code of the push request isn't the expected one

froodian:
Fixes #56.

This pull makes web-push return the pushResponse information from the push server.  This is important particularly for GCM where important information like a missing gcm api key, or invalid registration endpoint, are returned as part of the response body.

### DIFF
--- a/test/tests/testSendNotification.js
+++ b/test/tests/testSendNotification.js
@@ -170,6 +170,8 @@ suite('sendNotification', function() {
     .then(function() {
       assert(false, 'sendNotification promise resolved');
     }, function() {
+      assert(err instanceof webPush.WebPushError, 'err is a WebPushError');
+      assert(err.statusCode, 404);
       assert(true, 'sendNotification promise rejected');
     });
   });
--- a/test/tests/testSendNotification.js
+++ b/test/tests/testSendNotification.js
@@ -169,7 +169,7 @@ suite('sendNotification', function() {
     })
     .then(function() {
       assert(false, 'sendNotification promise resolved');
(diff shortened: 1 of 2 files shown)


marco-c:
Thanks! Could you use the custom Error class that I've introduced in 8c66db8b6581dc7c3faf6fe82d55f6f40ddb4484?
Simply `reject(new WebPushError('some message', pushResponse.statusCode))`.

marco-c:
It'd be great if you could also update the `test/tests/testSendNotification.js` test to check the error (check that the value with whom the promise is rejected is an instanceof WebPushError and that the status code is the expected one), but I can do it in a follow-up.

test/tests/testSendNotification.js
```diff
@@ -170,6 +170,8 @@ suite('sendNotification', function() {
     .then(function() {
       assert(false, 'sendNotification promise resolved');
     }, function() {
```
marco-c: You need to add `err` here :smile:

froodian: derp - sorry, didn't know how to run these tests locally, figured I'd let travis help me out ;)

froodian: fixed

marco-c: Right, they're quite cryptic (I've had to use a lot of hacks to write them). I've filed #67 to write the documentation.

marco-c:
Thanks again!

PR #69, opened 2015-12-23
Include response headers too

froodian:
GCM uses the Retry-After header during outages - http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.37


marco-c:
Thanks. Hopefully browsers will soon fully support the Web Push standard (https://tools.ietf.org/html/draft-ietf-webpush-protocol-02), so these kinds of errors will be clearly defined.

PR #70, opened 2016-01-27
Add new GCM Endpoint

nevi-me:
Google changed the endpoint to `https://gcm-http.googleapis.com/gcm/send` [here](https://developers.google.com/cloud-messaging/). My PR adds a check for the new endpoint.

I didn't touch the tests as I did this directly from Github. I can change the current endpoint to the new one if you'd like.

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,7 +72,7 @@ function sendNotification(endpoint, TTL, userPublicKey, payload) {
     }
 
     var gcmPayload;
-    if (endpoint.indexOf('https://android.googleapis.com/gcm/send') === 0) {
+    if (endpoint.indexOf('https://android.googleapis.com/gcm/send') === 0 || endpoint.indexOf('https://gcm-http.googleapis.com/gcm/send' === 0)) {
       if (payload) {
         reject(new WebPushError('Payload not supported with GCM'));
         return;


marco-c:
I'm still seeing the old URL in the latest Chromium version. Maybe they've changed it only for Android?

@beverloo do you know?

beverloo:
We don't plan to change the URL of the `endpoint` in Chrome until we switch to the Web Push protocol, so it's not needed to support Chrome.

marco-c:
> We don't plan to change the URL of the endpoint in Chrome until we switch to the Web Push protocol, so it's not needed to support Chrome.

Thank you for the info!

I guess we don't need to merge this PR then. Thank you anyway, @nevi-me!

nevi-me:
I also noticed after the PR that I'm still getting the old endpoint. I'll use both on my side so I don't miss the change when it happens. THanks @marco-c and @beverloo

marco-c:
@nevi-me when Chrome will switch to the Web Push protocol, there will be no difference between the Firefox and the Chrome code. That branch (the GCM branch) is only for the endpoints that use the current GCM protocol.

dvas0004:
It seems the Web Push change has happened though? I've had some difficult sending encrypted payloads, until I changed over to the new endpoint. Seems corroborated by Google's npm library web-push-encryption where they do a search and replace of the endpoints:

[https://github.com/GoogleChrome/web-push-encryption/blob/19a91f5c29ddd8dde7bd186fa67c9bf1548a2def/src/push.js#L69](https://github.com/GoogleChrome/web-push-encryption/blob/19a91f5c29ddd8dde7bd186fa67c9bf1548a2def/src/push.js#L69 )

beverloo:
To provide a quick update here, Chrome will support the Web Push protocol from Chrome 52 onward, but it will still require the developer to opt-in by supplying an `applicationServerKey` (VAPID) when subscribing for Push Messaging.

The endpoint will start with `https://fcm.googleapis.com/fcm/`.

dvas0004:
Thanks for the update! :)

Best Regards,
David Vassallo

*--------------------------------*
*@dave_vassallo <https://twitter.com/dave_vassallo>*
*http://blog.davidvassallo.me <http://blog.davidvassallo.me>*

PR #74, opened 2016-01-29
Fix payload with unicode characters

marco-c:


marco-c:
There's a failure downloading the ChromeDriver file, but tests are passing.

PR #82, opened 2016-02-25
Expect failure in encrypting/decrypting empty message on Node.js v0.12

marco-c:


marco-c:
`selenium-webdriver` isn't compatible with 0.12 anymore.

PR #84, opened 2016-02-26
JSON parse the body response

kylebuch8:
go ahead and parse the body so you don’t need to later.


marco-c:
The push service won't always reply with a JSON message (it isn't specified in https://tools.ietf.org/html/draft-ietf-webpush-protocol-03), so we can't call `JSON.parse` unconditionally.

If you want, we can parse the response as JSON when the `Content-Type` of the reply is `application/json`.

marco-c:
Please reopen if you still want to do it.

PR #90, opened 2016-03-08
Update to http_ece 0.4.1 and set explicit padding size

martinthomson:


martinthomson:
BTW, I tried running tests, but got this:

```
> node --harmony node_modules/.bin/istanbul cover _mocha -- --ui tdd test/tests/

Unable to resolve file [_mocha]
Try "istanbul help" for usage
npm ERR! Test failed.  See above for more details.
```

Probably some missing dependencies and commands.

marco-c:
> Probably some missing dependencies and commands.

I think you need to install `mocha` globally.

martinthomson:
> I think you need to install mocha globally.

I figured that was the case, but isn't that better put in package.json?

marco-c:
> I figured that was the case, but isn't that better put in package.json?

Yes, I'll do that. It was a legacy from the early days where I simply run tests using the `mocha` command instead of `npm test`.

PR #112, opened 2016-03-16
Implement VAPID

### LINKED ISSUE #79: Add VAPID functionality
Add optional functionality to append [VAPID](https://tools.ietf.org/html/draft-thomson-webpush-vapid-02) to push messages sent by web-push to push servers.

marco-c:
Fixes #79.

Still working on it, need to:
- [x] fix the interface (currently for testing I'm creating the keys every time you send a notification);
- [ ] fix sending a notification with payload with vapid;
- [x] add unit tests and selenium tests;
- [x] add documentation.


marco-c:
Waiting on https://github.com/mozilla-services/autopush/issues/410.

marco-c:
There's currently no way to support VAPID & payload with Firefox 45, so I'm not sure what to do. Maybe throw an exception when the user wants to use both? Or simply not support VAPID altogether until Firefox 46 is released.

marco-c:
For now, VAPID will be enabled only for Firefox and when sending a notification without payload.

PR #115, opened 2016-03-18
Support push payloads with GCM

marco-c:
Fixes #115.

Depends on https://github.com/martinthomson/encrypted-content-encoding/issues/17.


froodian:
Thanks for doing this so quickly, I look forward to trying it out within the next few days

PR #133, opened 2016-03-23
Adds command line executable

arcturus:
vapid and userAuth parameters configurable.

For using vapi add parameter --vapid.
For adding userAuth add parameter --auth=<your auth>

Example:
```bash
web-push --endpoint=https://updates.push.services.mozilla.com/push/v1/gAAAAABW8oWpBwhTzvvSxB-OKwsETosMvOsV75xUGLOQKXckw__twjoNHMIc-OkDljcs2aUaimxjhWUgZRcNZomdzdsgxXlW4dMpCiLzKa72JsQoEfBygiCHlmxxHlPJlimmzFSuk7eP --key=" BHwue1TPjc0sL4tnNway5HnW6qdD9Efi8AgUNyOw0CdUcNe2EvssLlqG8mK6Btc0t3KW9OCgSdJ6RQD6cpouzkY=" --payload="{\"body\":\"Molo mucho\"}"
```

### DIFF
--- a/bin/web-push.js
+++ b/bin/web-push.js
@@ -1,10 +1,11 @@
 #! /usr/bin/env node
+const fs = require('fs');
 const webPush = require('web-push');
 webPush.setGCMAPIKey(process.env.GCM_API_KEY);
 
 const argv = require('minimist')(process.argv.slice(2));
 
-const usage = 'Use: web-push --endpoint=<url> --key=<browser key> [--auth=<auth secret>] [--ttl=<seconds>] [--payload=<message>] [--vapid]';
+const usage = 'Use: web-push --endpoint=<url> --key=<browser key> [--auth=<auth secret>] [--ttl=<seconds>] [--payload=<message>] [--vapid-audience] [--vapid-subject] [--vapid-pvtkey] [--vapid-pubkey]';
 
 if (!argv['endpoint'] || !argv['key']) {
   console.log(usage);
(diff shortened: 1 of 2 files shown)


bin/web-push.js
```diff
@@ -0,0 +1,45 @@
...
+const payload = argv['payload'] || '';
+const auth = argv['auth'] || null;
+const useVAPID = argv['vapid'] || false;
+
+var params = {
+  TTL: ttl,
+  payload,
+  userPublicKey: key
+};
+if (useVAPID) {
```
marco-c: We should make the VAPID info configurable, so the user can define their own `audience` and `subject` and provide their keys.
We can do it in a follow-up.

bin/web-push.js
```diff
@@ -0,0 +1,45 @@
+#! /usr/bin/env node
+const webPush = require('web-push');
```
marco-c: Shouldn't this be `require('../index.js')`, so we always import the current version of the lib?

marco-c: Otherwise it would fail if you don't have the package installed globally.

arcturus: Actually I made it with that intention.

The use case is the following: I want to test my push notifications in the client and don't have the server side yet.

Solution:
You can do npm install -g web-push, and will provide you an executable to interact with the push service via command line

Does that make sense for you?

marco-c: If we use `require('../index.js')`, it would work both when installed locally and globally, no?

arcturus: Will try, in that case better to go with your suggestion.

arcturus:
Making vapid configurable, there are 4 new parameters:
--vapid-audience
--vapid-subject
If just those two are used, the private and public key will be autogenerated. In case you want to specify the keys you can use the following parameters (that should be locations to files in you file system):
--vapid-pvtkey
--vapid-pubkey

@marco-c  r?

PR #135, opened 2016-03-23
Make it possible to define VAPID keys to use during tests via env variables

### LINKED ISSUE #129: Use the same VAPID key for all Selenium tests
This way we can see statistics on the Mozilla Push Dashboard.

marco-c:
Fixes #129.


marco-c:
I've defined the env variables on Travis.

PR #159, opened 2016-04-30
Added support for chrome multicast notifications (payload not supported)

Xzya:
This adds support for sending a multicast notification to chrome. The endpoints argument is an array containing at least one endpoint, and at most 1000 endpoints. The endpoints can either be the full endpoints ("https://android.googleapis.com/gcm/send/" + subscriptionId), or just the subscriptionId.

This is useful for sending a large amount of notifications in a short time. While this doesn't support payloads, it's easy to add a callback to your server to get the payload in the service worker.

### DIFF
--- a/index.js
+++ b/index.js
@@ -284,10 +284,100 @@ function sendNotification(endpoint, params) {
   });
 }
 
+function sendChromeMulticastNotification(endpoints, params) {
+  var args = arguments;
+
+  return new Promise(function(resolve, reject) {
+    try {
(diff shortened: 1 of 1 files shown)


marco-c:
Unfortunately this is a GCM-only feature and isn't part of the Web Push standard.
Chrome is currently using the GCM protocol, but will soon switch to the Web Push standard, so I'd rather not support a non-standard method that will be removed sooner or later.

PR #162, opened 2016-05-27
Refactor of tests, Firefox + Chrome

gauntface:
This refactors the tests to be a little more maintainable and works on adding support for Chrome stable and beta in the tests.

Also simplified enabling / disabling tests in tests-selenium.js


gauntface:
Looks like chrome isn't running, @marco-c do you have a gcm api key in there travis environment?

Note: need to undo my manifest change

marco-c:
The env variable containing the GCM key was private, so it wasn't defined for external pull requests. I've now made it public.

.travis.yml
```diff
@@ -1,6 +1,15 @@
+sudo: required
```
marco-c: Nit: this is already at the end of the file, but is it still required now that you've removed line 26?

test/data/demo/manifest.json
```diff
@@ -3,6 +3,6 @@
   "short_name": "webpushdemo",
   "start_url": "/index.html",
   "display": "standalone",
-  "gcm_sender_id": "829371921996",
+  "gcm_sender_id": "653317226796",
```
marco-c: I guess this is your sender ID? Can you revert it to the old one?

.travis.yml
gauntface: Sorry didn't spot that, will remove it.

marco-c:
Looks like the tests are failing with Chrome.

gauntface:
Yeah I swapped out the gcm sender id when testing with my own good api key, so need to swap it back

gauntface:
Just squashed commit fixing the comments you raised. Tests should be good on Stable + Beta Chrome.

gauntface:
And of course with that, one test failed

gauntface:
I'll take a look at getting some more info out of these scenarios. I had one where gcm api returned an error but The tests weren't reading the response, otherwise might be useful to execute scripts to make the retrieval of the subscription and sending the push message.

If you can rerun the failing test it may pass (it did before the squash, but I'm worried about flakey tests)

marco-c:
Thanks! I'm going to merge this as it makes tests easier to improve, we can fix intermittent failures in the future.

PR #169, opened 2016-06-07
Fixes issue where getKeys would always return new VAPID keys

### LINKED ISSUE #168: web-push binary won't read public key and private key for VAPID
the if block uses variable names that are not declared :\

PaulKinlan:
Fixes #168 - the variable names were never actually declared so private keys and public keys from the command line would never be used.

### DIFF
--- a/bin/web-push.js
+++ b/bin/web-push.js
@@ -24,10 +24,10 @@ const vapidPvtKey = argv['vapid-pvtkey'] || null;
 
 function getKeys() {
   if (vapidPubKey && vapidPvtKey) {
-    const publicKey = fs.readFileSync(argv['vapid-pubkey']);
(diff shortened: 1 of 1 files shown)


gauntface:
The failing tests are from Firefox stable, we don't have marionette support at the moment (and even then I don't think it would work).

Would be good to add some tests here as well, but if you don't have time, I'm happy to write them after PR lands.

Generally LGTM from me. @marco-c any thoughts?

marco-c:
Yes, Selenium is currently broken with Firefox (it should hopefully be fixed soon: https://github.com/SeleniumHQ/selenium/issues/2110#issuecomment-224413609 and https://bugzilla.mozilla.org/show_bug.cgi?id=1278605).

marco-c:
We could keep using Firefox 46 for now.

PR #175, opened 2016-06-20
Remove support for the old encryption standard (retired in Firefox 45) and remove the old deprecated interface of the sendNotification function.

marco-c:


gauntface:
LGTM.

Issue #174, opened 2016-06-15
TTL not working


nikhil32:
Hi,

I am trying to include TTL key in sendNotification call with a time of 10 seconds. I closed my browser for 1 minute and then I reopened the browser and I can able to see my notification. I am not sure whether I am making something wrong. 

Thanks in advance....

marco-c:
Related: http://stackoverflow.com/questions/37853024/ttl-feature-not-working-on-google-chrome-push-notification

The problem is that the library supports the TTL for standard web push endpoints but not for GCM ones. I'm not sure if it's worth it to add support for TTL to GCM, since Chrome is soon going to use the standard web push protocol.

@gauntface what do you think?

nikhil32:
@marco-c that question was posted by me only. No one replied for that as well.

gauntface:
I'll raise a PR later this week to move Chrome behaviour over to web push. Sorry I've been delayed doing this as I've been moving from Europe to the States.

nikhil32:
Thanks for your inputs

marco-c:
This will be fixed by #180.

Issue #125, opened 2016-03-22
Remove support for the GCM protocol when it's no longer needed


marco-c:
I'm not sure when the standard Web Push service for Chrome is going to be available, but when it is, we should remove support for the GCM proprietary protocol.

gauntface:
I'll put together a test for Chrome's vapid -> web push protocol support so you can see how it'll work and see when gcm is no longer needed.

gauntface:
Sooooo I've put together the initial work for this BUT there is a little problem.

I can't get the VAPID JWT signing to work with the web-push library.

If you checkout simple-push-demo I've got it working: https://github.com/gauntface/simple-push-demo/blob/master/src/scripts/encryption/encryption-factory.js#L438

You can see where I'm current at here: https://github.com/gauntface/web-push/commit/f245bd839c378bfb58dec12780254d3617940951

The only thing I can think of is that the private key to PEM conversion is wrong. If you compare jwkToPem vs toPEM methods, you get quite different results, neither of which works.

Try running nodemon ./test/helpers/cli-run-server.js and copy the URL it gives you into Chrome Beta or Canary and you'll be able to see the endpoint change to FCM, which requires VAPID and uses web push protocol, but no matter what I do, I can't get a message sent. Any thoughts / ideas?

marco-c:
Interesting... It does work correctly with the Firefox push service (which does verify the token as far as I know). Maybe there's some difference in the VAPID implementation between GCM and Autopush?

marco-c:
@gauntface do you know if Opera and Samsung Internet Browser still require GCM?

froodian:
I did a personal assessment the other day where [it appeared to me](http://www.opera.com/docs/history/) that Opera updated to a version of Chromium that should support VAPID (52) for opera 39 (8/2016). It seems like maybe Samsung [did as well](https://medium.com/samsung-internet-dev/samsung-internet-v6-2-now-stable-ab7f95ed8b4b) in 10/2018? Of course, a non-trivial percentage of those browsers' users are not on those versions yet (https://caniuse.com/usage-table).

gauntface:
cc @torgo do you know who could confirm Samsung support.

theswerd:
Can someone explain to me why the GCM key is necessary in the first place — (Im new to notification sending)

froodian:
[...]

PR #180, opened 2016-06-25
Moving GCM endpoint over to using web push protocol

gauntface:
This removes the GCM protocol use. It does leave in the Authorization header needed for the GCM API Key though.

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,8 +115,6 @@ function sendNotification(endpoint, params) {
         }
       }
 
-      const isGCM = endpoint.indexOf('https://android.googleapis.com/gcm/send') === 0;
-
       var urlParts = url.parse(endpoint);
       var options = {
         hostname: urlParts.hostname,
@@ -143,26 +141,13 @@ function sendNotification(endpoint, params) {
         requestPayload = encrypted.cipherText;
       }
 
+      var isGCM = endpoint.indexOf('https://android.googleapis.com/gcm/send') === 0;
       if (isGCM) {
         if (!gcmAPIKey) {
           console.warn('Attempt to send push notification to GCM endpoint, but no GCM key is defined'.bold.red);
         }
 
-        var endpointSections = endpoint.split('/');
-        var subscriptionId = endpointSections[endpointSections.length - 1];
-
-        var gcmObj = {
-          registration_ids: [ subscriptionId ],
-        };
-        if (requestPayload) {
-          gcmObj['raw_data'] = requestPayload.toString('base64');
-        }
-        requestPayload = JSON.stringify(gcmObj);
-
-        options.path = options.path.substring(0, options.path.length - subscriptionId.length - 1);
-
         options.headers['Authorization'] = 'key=' + gcmAPIKey;
-        options.headers['Content-Type'] = 'application/json';
       }
 
       if (vapid && !isGCM) {
@@ -188,7 +173,7 @@ function sendNotification(endpoint, params) {
         options.headers['Authorization'] = 'Bearer ' + jwt;
         var key = 'p256ecdsa=' + urlBase64.encode(vapid.publicKey);
         if (options.headers['Crypto-Key']) {
-          options.headers['Crypto-Key'] += ',' + key;
+          options.headers['Crypto-Key'] += ';' + key;
         } else {
           options.headers['Crypto-Key'] = key;
         }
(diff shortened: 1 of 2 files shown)


gauntface:
Fixes #174 and #125 hopefully

index.js
```diff
@@ -143,28 +141,6 @@ function sendNotification(endpoint, params) {
...
-        }
-        requestPayload = JSON.stringify(gcmObj);
-
-        options.path = options.path.substring(0, options.path.length - subscriptionId.length - 1);
-
-        options.headers['Authorization'] = 'key=' + gcmAPIKey;
-        options.headers['Content-Type'] = 'application/json';
-      }
-
       if (vapid && !isGCM) {
```
marco-c: isGCM is used before its definition here.

index.js
```diff
@@ -188,7 +164,7 @@ function sendNotification(endpoint, params) {
         options.headers['Authorization'] = 'Bearer ' + jwt;
         var key = 'p256ecdsa=' + urlBase64.encode(vapid.publicKey);
         if (options.headers['Crypto-Key']) {
-          options.headers['Crypto-Key'] += ',' + key;
+          options.headers['Crypto-Key'] += ';' + key;
```
marco-c: Shouldn't this be a comma?
At least this is what Autopush is expecting: https://github.com/mozilla-services/autopush/blob/f546ed78585976d33534595dafcef401c118ea43/autopush/crypto_key.py and https://github.com/mozilla-services/autopush/blob/4361bbe07ee4b962622a3c16ddf632405a599158/autopush/tests/test_cryptokey.py

gauntface: If I put in a comma with GCM, it actually broke the GCM API and discussing this with @beverloo he pointed out:

 > Semi-colon separates fields in a single header value, comma separates header values.

which I cant find in a spec anywhere for semicolon but regarding comma [from this spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2):

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

This *might* have been one of the issues I was having with VAPID support on Chrome.

marco-c: Yes, semicolons separate parameters, commas separate values.

From http://httpwg.org/http-extensions/draft-ietf-httpbis-encryption-encoding.html#crypto-key:
> The Crypto-Key header field uses the extended ABNF syntax defined in Section 1.2 of [RFC7230]

From https://tools.ietf.org/html/rfc7230#section-1.2:
> This specification uses the Augmented Backus-Naur Form (ABNF) notation of [RFC5234] with a list extension, defined in Section 7, that allows for compact definition of comma-separated lists [...]

Then, in https://tools.ietf.org/html/draft-ietf-webpush-vapid-00#section-4:
> Note that with push message encryption [I-D.ietf-webpush-encryption], this results in two values in the Crypto-Key header field, one with the a "p256dh" key and another with a "p256ecdsa" key.

So, in this case, I think it should be a comma.

gauntface: My only counter argument is:

> one with the a "p256dh" key and another with a "p256ecdsa" key

They are are two (key, value) pairs which I'd view as two parameters. This is confused by:

>  this results in two values

Seems like a spec ambiguity @beverloo + @martinthomson

beverloo: /cc @martinthomson 

Which would, in turn, violate [draft-ietf-webpush-encryption](https://tools.ietf.org/html/draft-ietf-webpush-encryption-02#section-3.2):

> An application server MUST include exactly one entry in each of the Encryption and Crypto-Key header fields.

Semantically, having them in different entries makes sense because `p256ecdsa` has a different intended consumer (the push service) than `dh` (the user agent). This is unfortunately not compatible with all of today's implementations. FCM and Chrome enforce the single-entry restriction set by webpush-encryption, and will fail to deliver the message.

tl;dr— I suggest using `;` for compatibility reasons until we [fix the spec](https://github.com/webpush-wg/webpush-encryption/issues/6) and know the implementations can deal with this.

marco-c: From the implementation of `Autopush` it looks like, even if it assumed different values to be separated by comma, it doesn't break if you use a semicolon instead.
@jrconlin tested and confirmed this, so I think we can use the semicolon for the time being.

martinthomson:
FWIW, comma was the intent.  The semi-colon is probably fine.

PR #183, opened 2016-06-28
Moving Chrome to using VAPID

gauntface:
@marco-c any chance of a review?

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,6 +37,46 @@ function generateVAPIDKeys() {
   };
 }
 
+function getVapidHeaders(vapid) {
+  if (!vapid.audience) {
+    throw new Error('No audience set in vapid.audience');
+  }
+
+  if (!vapid.subject) {
+    throw new Error('No subject set in vapid.subject');
+  }
+
+  if (!vapid.publicKey) {
+    throw new Error('No key set vapid.publicKey');
+  }
+
+  if (!vapid.privateKey) {
+    throw new Error('No key set in vapid.privateKey');
+  }
+
+  var header = {
+    typ: 'JWT',
+    alg: 'ES256'
+  };
+
+  var jwtPayload = {
+    aud: vapid.audience,
+    exp: Math.floor(Date.now() / 1000) + 86400,
+    sub: vapid.subject,
+  };
+
+  var jwt = jws.sign({
+    header: header,
+    payload: jwtPayload,
+    privateKey: toPEM(vapid.privateKey),
+  });
+
+  return {
+    Authorization: 'Bearer ' + jwt,
+    'Crypto-Key': 'p256ecdsa=' + urlBase64.encode(vapid.publicKey)
+  };
+}
+
 function WebPushError(message, statusCode, headers, body) {
   Error.captureStackTrace(this, this.constructor);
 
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -152,30 +192,15 @@ function sendNotification(endpoint, params) {
 
       if (vapid && !isGCM) {
         // VAPID isn't supported by GCM.
+        vapid.audience = urlParts.protocol + '//' + urlParts.hostname;
```
marco-c: I'm a little confused here. Shouldn't the `audience` be the origin URL of the sender?

I wrote this code following https://blog.mozilla.org/services/2016/04/04/using-vapid-with-webpush/, but now reading the specification (https://tools.ietf.org/html/draft-thomson-webpush-vapid-02) it seems like `audience` should be the origin of the push resource URL...

@jrconlin

gauntface: I was very confused by the same thing.

> An "aud" (Audience) claim in the token MUST include the unicode
> serialization of the origin (Section 6.1 of [RFC6454]) of the push
> resource URL.  This binds the token to a specific push service.

Everyone I've spoken to regarding this matter thought it was the origin of the application server and after reading the spec wasnt sure what it should or could be. It's definitely one of those moments where the spec is phrased awkwardly, but FCM will reject and the final sentence in the spec regarding audience does suggest it should be the push service

jrconlin: @marco-c It's a bit complicated, but easier if you remember what VAPID is supposed to be. VAPID is basically a way for the feed provider to offer information about what the feed is. "Sub" is the critical item, and should be an email that Push Service providers can use to contact someone from the feed provider side. So it's the address you send the "Hey, we noticed a problem..." or "HOLY CRAP WOULD YOU CUT IT OUT" sort of messages. Since it's more than possible that one subscription service could provide lots of different subscriptions (Think of a proxy service like OneSignal or Feedly or someone) "Aud" is there to identify the individual feed. 

For that, it could be anything like the simple host origin, or a proxied ID URI, or honestly anything that lets the two operational teams quickly identify the feed that's having a problem. 

VAPID is basically an agreement between the feed provider and the push service. I'll note that the feed provider could also be a jerk and load VAPID with bogus information. In that case, the Push Service is completely within it's rights to deny the feed, ignore it, or rate limit it as much as they desire. VAPID is there pretty much for the benefit of the subscription provider.

So, in short: Yes, "aud" should contain a URL. What that URL actually is should be up to the subscription provider, but it'd be darn useful if it was meaningful in some fashion.

Now, all that said, I'll go look over the spec and see if we need to modify things.

gauntface: Thanks for the explanation @jrconlin.

In this case, this PR is a bad change and audience be defined externally from this library and it just so happens for FCM, the feed has to be for FCM and for mozilla, it can be the original url used. @marco-c does that sound good to you?

jrconlin: Also filed https://github.com/webpush-wg/webpush-vapid/issues/23 so that the spec gets clarified. I agree with @marco-c that it's not quite clear what the value or intent is in the spec. 

Well, that and apparently I failed to be properly caffeinated or focused when I read that the first time.

marco-c: @gauntface sorry, could you rephrase?

gauntface: Sorry, that was poorly phrased.

Basically, I'm thinking of changing the PR so that this can be defined externally (i.e. an option passed into sendNotification), like it was before and for the tests, I'll manually check Firefox or chrome and set the audience accordingly.

This does open the question of how should a developer find out what the correct audience is for a push service? Let's say a new browser Y started support push, the audience should be....?

marco-c: > This does open the question of how should a developer find out what the correct audience is for a push service? Let's say a new browser Y started support push, the audience should be....?

Maybe let's wait for https://github.com/webpush-wg/webpush-vapid/issues/23 to be addressed and then we can decide how to proceed.

jrconlin: +1 waiting. 

FWIW, being a push operator and grokking how customers might want to use VAPID, I'd want to be able to specify a URI of my own crafting for "aud", with the default being no value specified. Push server doesn't really care aside from trying to identify the reported source if the "sub" doesn't answer their emails, and as I noted, there's good reason why a subscription service might want to specify a different or seemingly arbitrary value.

marco-c:
https://github.com/webpush-wg/webpush-vapid/issues/23 has been addressed, so we can move forward with this.

marco-c:
Can you remove `audience` from the README, since it's no longer a parameter?

index.js
```diff
@@ -152,30 +192,15 @@ function sendNotification(endpoint, params) {
 
       if (vapid && !isGCM) {
```
marco-c: Is VAPID working now with GCM? Do we still need any GCM-specific code?

gauntface: In the time being this is still needed.

The GCM endpoint expects the authorization header to be the GCM Server API Key rather the the VAPID Bearer header.

GCM will continue to remain a thing for a few reasons: VAPID is only supported in Chrome Beta, but it's also the endpoint returned for Opera and Samsung Browser. I'm getting in touch with peeps at Opera and Samsung to see what the plan is, but it seems for the short term we should keep this in.

gauntface: One thing I didn't make clear as well - when you subscribe passing in the applicationSeverKey in Chrome, you get an FCM endpoint rather than a GCM endpoint, this is what causes isGCM to be false resulting in this code path, allowing FCM to authenticate the applicationServerKey

Issue #192, opened 2016-07-23
Change signature of sendNotification?


gauntface:
At the moment the signature of sendNotification results in code along
the lines of:

```javascript
const webpush = require('web-push');

webpush.setGCMAPIKey(GCM_API_KEY);

const subscriptions = getSubscriptions();
subscriptions.forEach(subscription => {
  const params = {
      payload: 'some payload',
      vapid: {
        subject: 'mailto: myemail@myemail.com',
        publicKey: VAPID_KEYS.publicKey,
        privateKey: VAPID_KEYS.privateKey
      }
  };

  if (subscription.keys) {
      params.userAuth = subscription.keys.auth;
      params.userPublicKey = subscription.keys.p256dh;
  }

  webpush.sendNotification(subscription.endpoint, params);
});
```

This could be simplfied to:

```javascript
const webpush = require('web-push');

webpush.setGCMAPIKey(GCM_API_KEY);
webpush.setVapidDetails({
    subject: 'mailto: myemail@myemail.com',
    publicKey: VAPID_KEYS.publicKey,
    privateKey: VAPID_KEYS.privateKey
})

const subscriptions = getSubscriptsion();
subscriptions.forEach(subscription => {
  webpush.sendNotification(subscription, 'some payload');
});
```

If we wanted to, we could move VAPID and GCM to be optional params
, as requested in #181:

```javascript
const webpush = require('web-push');

const subscriptions = getSubscriptsion();
subscriptions.forEach(subscription => {
  const options = {
    gcmAPIKey: getGCMAPIKey(subscription),
    vapid: getVAPIDDetails(subscription)
  }
  webpush.sendNotification(subscription, 'some payload', options);
});
```

What do you think @marco-c @nikhil32?

marco-c:
Both options look good to me. Maybe the first would be a smaller change compared to the second (it wouldn't change the signature for people who aren't using VAPID yet, which is the majority).

gauntface:
I was thinking both would be implemented but it would be up to the developer how to pass in GCM and VAPID - Matching PR #193.

Would you be ok with me putting together this change as a PR?

It would be a breaking change so may do a major release at a later stage, but I think it would be a good and worthwhile change.

marco-c:
> I was thinking both would be implemented but it would be up to the developer how to pass in GCM and VAPID - Matching PR #193.

> Would you be ok with me putting together this change as a PR?

OK, makes sense!

[...]

PR #193, opened 2016-07-23
Allow overriding GCM API key with a sendNotification argument.

### LINKED ISSUE #181: gcmAPIKey is a global variable
Hi,

I had a plan of hosting this web-push as a seperate rest server so that my java application will use it to send push notifications. I can have parallel threads running on java calling this server with different API keys. Since gcmAPIKey is a global variable, it can cause issues to me. How about having it also as a parameter in sendNotification as a optional overriding parameter. just a suggestion.

marco-c:
Fixes #181.

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,6 +131,8 @@ function sendNotification(endpoint, params) {
 
   return new Promise(function(resolve, reject) {
     try {
+      var curGCMAPIKey = gcmAPIKey;
+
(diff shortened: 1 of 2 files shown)


gauntface:
Will review later today. Any thoughts on #192?

gauntface:
Lgtm

Issue #207, opened 2016-08-30
Reduce size of install


gauntface:
See : https://twitter.com/borismus/status/770456100235726849 from @borismus

marco-c:
Maybe we could find a way to make the dependencies needed to support older Node.js versions optional.

marco-c:
Another idea is to create PR to our dependencies to make them add a `.npmignore` file to avoid including tests in their packages, e.g. https://github.com/indutny/asn1.js/pull/75.

marco-c:
With the new `http_ece` we're under 1 MB.

marco-c:
List of packages where more files could be added to .npmignore: `typedarray`, `minimist`, `jws`, `jwa`, `base64url`. We would save around 135 kB.

marco-c:
https://github.com/substack/typedarray/pull/11
https://github.com/substack/minimist/pull/95
https://github.com/brianloveswords/node-jws/pull/60
https://github.com/brianloveswords/node-jwa/pull/15
https://github.com/brianloveswords/base64url/pull/14

marco-c:
I think we can call this fixed!

PR #196, opened 2016-07-24
Adding stronger eslint - fixing linting issues, re-org of repo 

gauntface:
I've moved the source code into src/ for the module and fixed up a range of linting issues after moving to the airbnb eslint config.

Closes #194

PR #208, opened 2016-08-30
Remove reference to binary and no longer used minimist package

marco-c:
We've removed the binary in #196, but no one has complained. I think we can avoid to reintroduce it and save a bit of space (helping #207) by removing the `minimist` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,6 @@
   "version": "2.2.0",
   "description": "Web Push library for Node.js",
   "main": "src/index.js",
-  "bin": {
-    "web-push": "bin/web-push.js"
-  },
   "scripts": {
     "download-browser": "node --harmony ./test/helpers/download-test-browsers.js",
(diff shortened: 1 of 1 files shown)


ithinkihaveacat:
FWIW, I would have found `web-push.js` useful if I'd known it existed. (In my ignorance I wrote my own version. Was going to create a PR to add it here, but never got around to it.)

marco-c:
> FWIW, I would have found web-push.js useful if I'd known it existed. (In my ignorance I wrote my own version. Was going to create a PR to add it here, but never got around to it.)

What is your use case? We can still keep it, it doesn't really increase the size of the library so much.

gauntface:
Apologies for removing it :(

ithinkihaveacat:
@marco-c My use case was pretty much just testing and demos--when testing front-end stuff I wanted an easy way to send backend messages.

marco-c:
Closing, will reopen a new PR to reintroduce the file.

PR #220, opened 2016-09-20
Changing vapid input / output to be base64url encoded string

### LINKED ISSUE #210: Change output and input of VAPID from Buffer to String
I'm constantly hitting issues with messing with VAPID keys as it's a buffer that I need to get into the web page, convert to unit8array and then find a way to store the VAPID keys so that I can then get that value back out and pass into the web-push library.

Exporting and importing VAPID keys as base64 url safe strings at least makes storing and re-using easier with the web push libs and we can then provide an example of how to use this base64 url string as the applicationServerKey in the page.

@marco-c what do you think?

gauntface:
Fixes #210 and adds some tests to setVapidDetails().

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -55,35 +55,9 @@ WebPushLib.prototype.setVapidDetails =
       return;
     }
 
-    if (typeof subject !== 'string' || subject.length === 0) {
-      throw new Error('The subject should be a URL or mailto:');
-    }
-
-    if (subject.indexOf('mailto:') !== 0) {
-      // Must be a url if it doesn't have mailto at the start
-      const subjectParseResult = url.parse(subject);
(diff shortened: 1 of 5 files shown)


src/web-push-lib.js
```diff
@@ -55,6 +55,36 @@ WebPushLib.prototype.setVapidDetails =
       return;
     }
 
+    if (typeof subject !== 'string' || subject.length === 0) {
```
marco-c: Perhaps we can move the checks in a separate function in vapid-helper.js and use it everywhere, so we don't have duplicate code?

gauntface: Of course :+1: 

Changes done.

gauntface:
*nudge* @marco-c

PR #240, opened 2016-10-07
Remove unneeded keyid parameter.

### LINKED ISSUE #234: When is keyid=p256dh; needed?
I only noticed this today: https://github.com/web-push-libs/web-push/blob/master/src/web-push-lib.js#L167

        requestOptions.headers.Encryption = 'keyid=p256dh;salt=' +
          encrypted.salt;
        requestOptions.headers['Crypto-Key'] = 'keyid=p256dh;dh=' +
          urlBase64.encode(encrypted.localPublicKey);

I've not used the keyid anywhere in my simple push demo and just wondering when it would be needed? Looking at the spec for web push protocol the only reference I found was:

>   An application server MUST include exactly one entry in the
>   Encryption field, and at most one entry having a "dh" parameter in
 >  the Crypto-Key field.  This allows the "keyid" parameter to be
 >  omitted from both header fields.

marco-c:
Fixes #234.

Let's try.

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -164,10 +164,8 @@ WebPushLib.prototype.sendNotification =
         requestOptions.headers['Content-Length'] = encrypted.cipherText.length;
         requestOptions.headers['Content-Type'] = 'application/octet-stream';
         requestOptions.headers['Content-Encoding'] = 'aesgcm';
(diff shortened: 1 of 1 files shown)


gauntface:
LGTM

PR #241, opened 2016-10-07
Set max-len to 120 characters

marco-c:
@gauntface do you mind if we remove this rule? I don't like being forced to a maximum of 80 characters sometimes at the expense of readability.


gauntface:
Fine with me. Would you be open to extending it to something line 120 just to force some level of sanity?

BUT if you feel strongly then by all means leave it as is :)

marco-c:
120 sounds good.

PR #251, opened 2016-11-03
Adding headers key to options for supporting future headers.

### LINKED ISSUE #235: Add Urgency Support / More generic way to add additional headers?
I wasn't aware of the Urgency header until I noticed it in the web-push-php library.

https://tools.ietf.org/html/draft-ietf-webpush-protocol-10#section-5.3

Given that this requires extra support from us and issue #211 

I'm wondering if it makes sense to support adding arbitrary headers, something along the lines of:

```
webpush.sendNotification(subscription, payload, {
    headers:  {
        'Urgency': 'value',
        'Topic': 'value',
        'Future-Header': 'value'
    }
});
```

We can then support future headers without needing to do anything and we can add additional and checks if we want to.

Main downside is that it's exposing the implementation of web push protocol somewhat and we probably should move TTL into headers if we move forward with this.

The alternative is we grow the options object, but it will require explicit support for new headers.

```
webpush.sendNotification(subscription, payload, {
      'Urgency': 'value',
      'Topic': 'value',
      'Future-Header': 'value'
});
```

@marco-c What do you think?

delapuente:
Fix #235 

@marco-c or @gauntface do you mind to take a look at my patch, please?

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -106,9 +106,11 @@ WebPushLib.prototype.generateRequestDetails =
     let currentGCMAPIKey = gcmAPIKey;
     let currentVapidDetails = vapidDetails;
     let timeToLive = DEFAULT_TTL;
+    let extraHeaders = {};
 
     if (options) {
       const validOptionKeys = [
+        'headers',
         'gcmAPIKey',
         'vapidDetails',
         'TTL'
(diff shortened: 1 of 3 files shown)


marco-c:
LGTM. Could you also add a test in testSendNotification to check that the extra headers are actually sent?

I'd also add a way to define `Topic` and `Urgency` directly instead of as extra headers (like for `TTL`), but we can do it in a follow-up.

Please also update to master after #252 lands, so that we can see tests passing.

delapuente:
@marco-c or @gauntface, do you think we could merge this now?

marco-c:
Sorry I hadn't noticed you pushed new changes!

delapuente:
What a silly mistake. It should work now.

PR #256, opened 2016-11-22
Issue template

gauntface:
@marco-c What do you think of adding an issue template for this project? Would be good to ask some of this info up front and in a more consistent manner.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,25 @@
+# Setup
+
+> Please provide the following details, the more info you can provide the
+> better.
+
+*Operating System*: <Linux | Windows | OS X>
+*Node Version*: <4.2 | 6.0 | ...>
+*web-push Version*: <1.0.0>
+
+# Problem
+
+> Please explain what behaviour you are seeing.
+
+# Expected
+
+> Please explain what you expected to happen
+
+# Example / Reproduce Case
+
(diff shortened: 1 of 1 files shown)


marco-c:
Sounds good to me.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+> better.
+
+*Operating System*: <Linux | Windows | OS X>
+*Node Version*: <4.2 | 6.0 | ...>
+*web-push Version*: <1.0.0>
+
+> Please select any browsers that you are experiencing problems with:
+
+- [ ] Google Chrome
+- [ ] Firefox
```
marco-c: Nit: For consistency, let's change this to `Mozilla Firefox` or `Google Chrome` to `Chrome`.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+- [ ] Opera for Android
+- [ ] Samsung Internet Browser
+- [ ] Other
+
+> Have you tried different versions of these browsers? I.e. Google Chrome
+> Beta and Firefox Beta? Did they produce the same result?
```
marco-c: Can we also suggest to try with different browsers (in case the user has only tested with one)?
I'm pretty sure we would ask this anyway.

gauntface:
@marco-c Just added the changes you mentioned, could you take another look.

Issue #254, opened 2016-11-09
error when use chrome  54.0.2840.71 (64-bit)


fix2015:
Hi, try use your library, all did like you wrote, but on Firefox it work perfect with chrome 54.0.2840.71 (64-bit) i get error

` [WebPushError: Received unexpected response code]
  name: 'WebPushError',
  message: 'Received unexpected response code',
  statusCode: 400,
  headers: 
   { 'content-type': 'text/html; charset=UTF-8',
     date: 'Wed, 09 Nov 2016 13:38:11 GMT',
     expires: 'Wed, 09 Nov 2016 13:38:11 GMT',
     'cache-control': 'private, max-age=0',
     'x-content-type-options': 'nosniff',
     'x-frame-options': 'SAMEORIGIN',
     'x-xss-protection': '1; mode=block',
     server: 'GSE',
     'alt-svc': 'quic=":443"; ma=2592000; v="36,35,34"',
     'accept-ranges': 'none',
     vary: 'Accept-Encoding',
     connection: 'close' },
  body: '<HTML>\n<HEAD>\n<TITLE>UnauthorizedRegistration</TITLE>\n</HEAD>\n<BODY BGCOLOR="#FFFFFF" TEXT="#000000">\n<H1>UnauthorizedRegistration</H1>\n<H2>Error 400</H2>\n</BODY>\n</HTML>\n' `
Maybe you know where can be pb?

fix2015:
with Firefox i get this and all work 
`{ statusCode: 201,
  body: '',
  headers: 
   { 'access-control-allow-headers': 'content-encoding,encryption,crypto-key,ttl,encryption-key,content-type,authorization',
     'access-control-allow-methods': 'POST',
     'access-control-allow-origin': '*',
     'access-control-expose-headers': 'location,www-authenticate',
     'content-type': 'text/html; charset=UTF-8',
     date: 'Wed, 09 Nov 2016 13:38:23 GMT',
[...]

PR #258, opened 2016-11-28
VAPID Expiration within 12 Hours

gauntface:
May be a solution to #254

### DIFF
--- a/src/vapid-helper.js
+++ b/src/vapid-helper.js
@@ -123,7 +123,8 @@ function getVapidHeaders(audience, subject, publicKey, privateKey, expiration) {
   publicKey = urlBase64.decode(publicKey);
   privateKey = urlBase64.decode(privateKey);
 
-
+  const DEFAULT_EXPIRATION = Math.floor(Date.now() / 1000) + 43200;
+  
   if (expiration) {
     // TODO: Check if expiration is valid and use it in place of the hard coded
     // expiration of 24hours.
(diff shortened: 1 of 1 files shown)


marco-c:
Why does setting the expiration to 12 hours instead of 24 hours help?

Need to fix some linting errors.

gauntface:
The spec defines that 24 hours is the max expiration for a certificate. My understanding is that FCM is ensuring this isn't exceeded and failing if it does.

By defining the default expiration as 24 hours in the future, there is no margin for error with the machines local time. The result is that if the local time is out by a couple of minutes, you may exceed the 24 hour limit resulting in an invalid VAPID header.

By using 12 hours as the default, we are just introducing a margin of error if the local time is set in the future.

marco-c:
> The spec defines that 24 hours is the max expiration for a certificate. My understanding is that FCM is ensuring this isn't exceeded and failing if it does.

Ah, I see, I thought the failure was due to the expiration being in the past (for users who had a clock set to the past).

marco-c:
Once Travis is happy, I'm going to merge this and release a point release.

gauntface:
The clock error can be caused by both future and past. This change will change from 24hr margin of error for past and 0 margin of error for future to 12hr for both.

@p-nowak did comment that this solved the problem for one of his test cases, so I think it's a useful change. Obviously better error messages would help greatly with this.

marco-c:
I've released `3.2.1`.

PR #263, opened 2017-02-06
GCM Error no longer undefined

gauntface:
At the moment the error message shows up as undefined (see here for screenshot: https://github.com/web-push-libs/web-push/issues/237#issuecomment-277581487)


marco-c:
That's sneaky!

Could you add a reference to setGCMAPIKey (as you suggested in https://github.com/web-push-libs/web-push/issues/254#issuecomment-277752407)?

gauntface:
@marco-c just added a little more info to the log.

marco-c:
Released 3.2.2.

PR #264, opened 2017-02-09
Expose WebPushError as a public API

perrin4869:
This can be helpful in identifying errors, such as `err instanceof WebPushError`, etc. For your consideration!


marco-c:
Thanks!

perrin4869:
@marco-c not that it's anything urgent, but any clue when you might release this? Thanks!

marco-c:
I've just published a new version with the fix.

PR #266, opened 2017-02-10
Returned correct push endpoint in case of error.

shahrukh330:
Library was returning incorrect endpoint if webpush.sendNotification(....) was called in a loop.


marco-c:
Thanks!

PR #267, opened 2017-02-12
Replaced hostname with host

firefligher:
Google says in their developer pages
> This is the origin of the push service (NOT the origin of your site). In JavaScript, you could do the following to get the audience: const audience = new URL(subscription.endpoint).origin

[Source](https://developers.google.com/web/updates/2016/07/web-push-interop-wins#sending_a_push_message)

W3C says (at least in W3CSchools)
> Return the protocol, hostname and port number of a URL.

[Source](http://www.w3schools.com/jsref/prop_loc_origin.asp)

I guess, the old line would cause problems, if a browser vendor decided to run their endpoint service somewhere else than port 80 or 443.

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -205,7 +205,7 @@ WebPushLib.prototype.generateRequestDetails =
     } else if (currentVapidDetails) {
       const parsedUrl = url.parse(subscription.endpoint);
       const audience = parsedUrl.protocol + '//' +
-        parsedUrl.hostname;
+        parsedUrl.host;
 
       const vapidHeaders = vapidHelper.getVapidHeaders(
         audience,


gauntface:
Is there any chance you could add a test case that ensures the port number
is set (just to cover off any future changes reverting this)

firefligher:
Honestly, I don't think that I have the required knowledge to do that. I rarely use JS and I don't have a clue about test suits in Node.

gauntface:
Ok, I raised an issue to add a test for this. @marco-c I can't see any harm in this given the integration tests are passing and this would be the spec'd behavior.

abdulhannanali:
Although the behavior of origin in both cases will be the same. Just for the sake of clarity as the W3Schools link to `location.origin`  the actual origin property we should refer to in the browser is of the URL instance and it's documentation can be found here https://developer.mozilla.org/en-US/docs/Web/API/URL/origin

Issue #280, opened 2017-05-22
ETIMEDOUT in sendNotification.Can it add a proxy option?


yyh1102:
I've already used webpush.sendNotification and it automatically request the Google server address.But I received so many ETIMEDOUT. 
![image](https://cloud.githubusercontent.com/assets/10328169/26300413/9962a8fa-3f10-11e7-9631-a6d6524309a9.png)

For the server in China, the network is unluckily blocked by gov, so we have to use proxy server to link websites abroad. 

**Hope the web-push can add proxy options.**

marco-c:
What options would you need?

yyh1102:
Like
```
const options={
  ...,
  proxy:"http://xxx.xxx.xxx"  // my proxy server
}
webpush.sendNotification(
  pushSubscription,
  payload,
  options
);
```
So that the web-push can request the google url through my proxy server.

marco-c:
I'd happily accept a pull request for this.

You can also implement the same without using a standard proxy. Just change the endpoint URL to be the URL of your custom proxy (e.g. YOUR_PROXY_URL?endpoint=ACTUAL_ENDPOINT_URL) and make the proxy send the request to the endpoint URL.

yyh1102:
Oh,that's a good idea~ But actually the proxy server is not mine, and i can do nothing on it to build such custom api.
A pr can be ok~
Thanks for replying.

yyh1102:
Hey due. I've forked your repo and add proxy api. But after serious consideration, i think it's not necessary to pull request. There are three reasons:
- The demand of proxy is just for Chinese developers, so this api will be very duplicate for other people.
- The original code can not be run in my local environment for some reasons, so I can't do automatic test.
- For convenience I only change https to http to request the Push Server using proxy, and it works in my project. 

So it's better to make it an independent repo, and this is the [link](https://github.com/yyh1102/web-push-China)

MartijnDwars:
Note that it's not only Chinese developers that are affected by this problem. If you are behind a corporate proxy, which is common in big organisations, you encounter the same issue. Having the ability to set a proxy, or better, respecting the HTTP(S)_PROXY environment variables, would benefit these users. That said, #282 provides exactly this functionality, so I think the focus should be on integrating that PR.

yyh1102:
Oh. I have not noticed that PR. It's really good if the official lib implements that API. Thanks for contributors!!

Peanoquio:
Hi there,
[...]

PR #282, opened 2017-06-06
adding request to support HTTPS_PROXY

refextu:
Not the most elegant way to solve https://github.com/web-push-libs/web-push/issues/280 but adds common used npm request to provide solid request options and HTTPS_PROXY env support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,8 @@
     "http_ece": "^0.5.2",
     "jws": "^3.1.3",
     "minimist": "^1.2.0",
+    "request": "^2.81.0",
+    "request-promise-native": "^1.0.4",
     "urlsafe-base64": "^1.0.0"
   },
   "devDependencies": {
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -2,7 +2,7 @@
 
 const urlBase64 = require('urlsafe-base64');
 const url = require('url');
-const https = require('https');
+const request = require('request-promise-native');
 
 const WebPushError = require('./web-push-error.js');
 const vapidHelper = require('./vapid-helper.js');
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -33,6 +33,8 @@
     "http_ece": "^0.5.2",
     "jws": "^3.1.3",
     "minimist": "^1.2.0",
+    "request": "^2.81.0",
+    "request-promise-native": "^1.0.4",
```
gauntface: Does request-promise-native not have it's own dependency on request?

gauntface: Also any reason not to use node-fetch for this?

src/web-push-lib.js
```diff
@@ -254,45 +254,32 @@ WebPushLib.prototype.sendNotification =
     }
 
     return new Promise(function(resolve, reject) {
```
gauntface: Can we remove this out promise since the new API is returning a promise?

gauntface:
Is there any chance of adding a basic test to ensure this works (i.e. set a HTTPS_PROXY environment variable to point to a local server and pass the test if the local server receives a request).

package.json
refextu: request-promise-native asks for peer dependency, added to avoid warnings

src/web-push-lib.js
refextu: yes, that should be possible, tried to change as little code as possible

marco-c:
@refextu ping?

refextu:
was in holiday, try to write a test this week, not so easy to setup a portable proxy under test conditions.

gauntface:
@refextu if you get stuck I'm more than happy to help (via chat, email, hangout or PR into your branch).

abdulhannanali:
@gauntface Can I pick this up?

gauntface:
That's up to @refextu

marco-c:
@abdulhannanali I suppose @refextu is not going to work on this anymore, so you can pick this up if you want.

Dhappi:
@marco-c not sure if anyone is working on this issue. If not can i pick it up ?

marco-c:
@Dhappi definitely! I'd love to see this finally completed.

Dhappi:
I tried the branch by @refextu and it seems to work fine with process.env.http_proxy. so do we just need a test? or do we need an options object for defining proxy too ?

marco-c:
> I tried the branch by @refextu and it seems to work fine with process.env.http_proxy. so do we just need a test? or do we need an options object for defining proxy too ?

A test and addressing @gauntface's review comments.

marco-c:
It's a pity that `request` has so many dependencies, it goes against our goal of having a small library.

Dhappi:
i can spend some time looking into an alternate solution to ```request```.

gauntface:
@marco-c is there a reason not to go with `node-fetch`? Generally I'm a big fan of heading in that direction - but I'm also unsure if it has fewer dependencies or not.

Dhappi:
@gauntface i have something working with [http-proxy-agent](https://www.npmjs.com/package/https-proxy-agent) without using request. I like the idea of using node-fetch but i think we would still need a proxy agent for that.

gauntface:
@marco-c shall we close this PR in favor of someone else filing a PR implementing the feature with tests?

marco-c:
Yeah, good idea.

Hotcapsicum:
> @refextu if you get stuck I'm more than happy to help (via chat, email, hangout or PR into your branch).
so , if I have proxy ,then I can use web-push in China? I need use proxy to request fcm

marco-c:
@Hotcapsicum there is an option to use a proxy with the library, just look at the options of the `sendNotification` function in the README: https://github.com/web-push-libs/web-push.

PR #286, opened 2017-07-28
Add Test For Port Number passed to Audience in Vapid

### LINKED ISSUE #270: Add test for port number passed to 'audience' in Vapid
This is a note to add test to ensure behavior from https://github.com/web-push-libs/web-push/pull/267 is kept.

abdulhannanali:
Integration test for Endpoint passed to Audience in Vapid. This uses jws to decode the Authorization header returned with the value and tests if it contains the expected audience.

This is a test to accompany the change in #267 and resolves #270.

### DIFF
--- a/test/test-generate-request-details.js
+++ b/test/test-generate-request-details.js
@@ -258,12 +258,17 @@ suite('Test Generate Request Details', function() {
       }
     };
 
-    const message = undefined;
-    const requestDetails = webPush.generateRequestDetails(subscription, message, extraOptions);
+    const requestDetails = webPush.generateRequestDetails(subscription, null, extraOptions);
     const authHeader = requestDetails.headers.Authorization;
-    const audience = jws.decode(authHeader.slice(8)).payload.aud;
 
-    assert.ok(audience, 'Audience does not exist');
-    assert.equal(audience, 'http://example.com:4242', 'Audience does not contain expected value');
+    // Get the Encoded JWT Token from the Authorization Header
+    // and decoded it using `jws.decode`
+    // to get the value of audience in jwt payload
+    const jwtContents = authHeader.match(/WebPush\s(.*)/)[1];
+    const decodedContents = jws.decode(jwtContents);
+    const audience = decodedContents.payload.aud;
+
+    assert.ok(audience, 'Audience exists');
+    assert.equal(audience, 'http://example.com:4242', 'Audience contains expected value with port');
   });
 });
--- a/test/test-generate-request-details.js
+++ b/test/test-generate-request-details.js
@@ -4,6 +4,7 @@ const assert = require('assert');
 const urlBase64 = require('urlsafe-base64');
 const webPush = require('../src/index');
 const crypto = require('crypto');
+const jws = require('jws');
 
 suite('Test Generate Request Details', function() {
   test('is defined', function() {
@@ -242,5 +243,27 @@ suite('Test Generate Request Details', function() {
     assert.equal(details.headers.TTL, extraOptions.TTL);
     assert.equal(details.headers.Topic, extraOptions.headers.Topic);
     assert.equal(details.headers.Urgency, extraOptions.headers.Urgency);
+  });
+
+  test('Audience contains port', function() {
+    const subscription = {
+      endpoint: 'http://example.com:4242/life-universe-and-everything'
+    };
+
+    const extraOptions = {
+      vapidDetails: {
+        subject: 'mailto:example@example.com',
+        publicKey: vapidKeys.publicKey,
+        privateKey: vapidKeys.privateKey
+      }
+    };
+
+    const message = undefined;
+    const requestDetails = webPush.generateRequestDetails(subscription, message, extraOptions);
+    const authHeader = requestDetails.headers.Authorization;
+    const audience = jws.decode(authHeader.slice(8)).payload.aud;
+
+    assert.ok(audience, 'Audience does not exist');
+    assert.equal(audience, 'http://example.com:4242', 'Audience does not contain expected value');
   });
 });


abdulhannanali:
@gauntface We'll be able to unit test this logic if we move the logic of generating the Audience value for the payload to a function. Is it preferable or does it work like this too?

abdulhannanali:
Thanks a lot for the opportunity to contribute. Web Push is awesome :heart:

abdulhannanali:
@gauntface The errors failing seem unrelated. Can you guide please? Seems to be failing on node4 mostly.

abdulhannanali:
Test Seems to be passing for Node 4. Check out details to see the output.

<details>
```

  Test CLI
    ✓ no args run (99ms)
    ✓ test send-notification no args (94ms)
    ✓ test send-notification only endpoint (270ms)
    ✓ test send-notification all options (105ms)
    ✓ test generate vapid keys (98ms)
    ✓ test generate JSON vapid keys (104ms)

  Test Encryption Helpers
    ✓ is defined
    ✓ encrypt/decrypt string
    ✓ encrypt/decrypt buffer
    ✓ bad input to encrypt

  Test Generate Request Details
    ✓ is defined
    ✓ 0 arguments
    ✓ No Endpoint
    ✓ Empty Endpoint
    ✓ Array for Endpoint
    ✓ Object for Endpoint
    ✓ Object for Endpoint
    ✓ Payload provided with no keys
    ✓ Payload provided with invalid keys
    ✓ Payload provided with only p256dh keys
    ✓ Payload provided with only auth keys
    ✓ userPublicKey argument isn't a string
    ✓ userAuth argument isn't a string
    ✓ userPublicKey argument is too long
    ✓ userPublicKey argument is too short
    ✓ userAuth argument is too short
    ✓ rejects when payload isn't a string or buffer
    ✓ send notification with invalid vapid option
    ✓ duplicated headers
    ✓ Extra headers
    ✓ Audience contains port

  setVapidDetails()
    ✓ is defined
    ✓ Valid URL input
    ✓ Valid mailto: input
    ✓ reset Vapid Details with null
    ✓ Invalid input should throw an error

  Test Vapid Helpers
    ✓ is defined
    ✓ generate vapid keys
    ✓ generate new vapid keys between calls
    ✓ should throw errors on bad input
    ✓ should get valid VAPID headers


  41 passing (813ms)

```
</details>

test/test-generate-request-details.js
```diff
@@ -243,4 +244,26 @@ suite('Test Generate Request Details', function() {
...
+  test('Audience contains port', function() {
+    const subscription = {
+      endpoint: 'http://example.com:4242/life-universe-and-everything'
+    };
+
+    const extraOptions = {
+      vapidDetails: {
+        subject: 'mailto:example@example.com',
+        publicKey: vapidKeys.publicKey,
+        privateKey: vapidKeys.privateKey
```
marco-c: These should be URL Safe Base64 Encoded.

test/test-generate-request-details.js
```diff
@@ -243,4 +244,26 @@ suite('Test Generate Request Details', function() {
...
+      }
+    };
+
+    const message = undefined;
```
marco-c: Nit: remove this and simply pass `null` as the second argument to `generateRequestDetails`

test/test-generate-request-details.js
```diff
@@ -243,4 +244,26 @@ suite('Test Generate Request Details', function() {
...
+    const requestDetails = webPush.generateRequestDetails(subscription, message, extraOptions);
+    const authHeader = requestDetails.headers.Authorization;
+    const audience = jws.decode(authHeader.slice(8)).payload.aud;
+
+    assert.ok(audience, 'Audience does not exist');
```
marco-c: For this and the following, please write what you are asserting instead of what makes the assertion fail (`Audience exists` instead of `Audience does not exist`).

test/test-generate-request-details.js
```diff
@@ -243,4 +244,26 @@ suite('Test Generate Request Details', function() {
...
+        subject: 'mailto:example@example.com',
+        publicKey: vapidKeys.publicKey,
+        privateKey: vapidKeys.privateKey
+      }
+    };
+
+    const message = undefined;
+    const requestDetails = webPush.generateRequestDetails(subscription, message, extraOptions);
+    const authHeader = requestDetails.headers.Authorization;
+    const audience = jws.decode(authHeader.slice(8)).payload.aud;
```
gauntface: Nit: Could you pull out the `slice` to a separate line and explain what you are doing or do something like:

```
const jwtContents = authHeader.slice('WebPush ');
```

gauntface:
@abdulhannanali thanks for the PR.

Looking into Travis one of the test matrix's is failing on your test. Might require using NVM to install node 6 and see if it reproduces locally. One of the other matrixs failed from a timeout - I restarted it to see if it reproduces.

test/test-generate-request-details.js
abdulhannanali: @marco-c These are taken from `vapidKeys` available in the parent closure of `suite` function call, and there they are generated using `generateVAPIDKeys` method of `vapid-helper.js` which already does this.

Since,  I went into their implementation and it's doing both of these things

```js
function generateVAPIDKeys() {
  const curve = crypto.createECDH('prime256v1');
  curve.generateKeys();

  return {
    publicKey: urlBase64.encode(curve.getPublicKey()),
    privateKey: urlBase64.encode(curve.getPrivateKey())
  };
}
```

test/test-generate-request-details.js
abdulhannanali: Neat thanks! :smile:

abdulhannanali:
@gauntface I tested this both on node4 and node6 using nvm and it passes for both. Don't have any idea why it failed. This passes for Node v6.7.1 but since It was tested using v6.11.1 I will run this test against this version too.

test/test-generate-request-details.js
abdulhannanali: I am decoding the JWT without verifying the signature, since it's coming from a trusted source, to get the decoded contents and then verify the `aud` value with the expected one.

abdulhannanali:
@gauntface @marco-c Thank you for reviewing the PR I have resolved some of the things. One solution to the error that occured on Travis CI but not on my local computer can be to use
values such as the ones below

```js
const VALID_PUBLIC_KEY = urlBase64.encode(new Buffer(65));
const VALID_PRIVATE_KEY = urlBase64.encode(new Buffer(32));
```

This excerpt is taken from test for vapid-helper.js but it should work otherwise too.

test/test-generate-request-details.js
marco-c: Right, I forgot it was already doing that.

It's unexpected then that it doesn't work. The failing test was saying that the key was not of the correct length, but it shouldn't be happening (we even have a test for this: https://github.com/web-push-libs/web-push/blob/master/test/test-vapid-helper.js#L29).

@gauntface any idea?

marco-c: @abdulhannanali can you try adding some logging to the test?
Log `vapidKeys.publicKey` and `vapidKeys.privateKey` and their decoded versions in the test, so we can see what they look like.

marco-c:
The test is still failing now, could you add some logging (https://github.com/web-push-libs/web-push/pull/286#discussion_r131036611)?

gauntface:
@abdulhannanali it would be good to get some console logging in there to see what the base encoded value comes out like.

I'm struggling to see why it's failing especially on just one version of node :-/

gauntface:
*apologies - did not mean to close.

marco-c:
I'm going to merge this since the tests appear to be intermittently failing and unrelated to this PR. I've filed #295 to track the intermittent failure.

abdulhannanali:
@gauntface Sorry just saw the comments right now. I think it's unrelated to the implementation here and tests appear to be failing intermittently as @marco-c suggested. I would love to work on solving the problems we have with test suites. Since this has been merged, is there still need of logging?

marco-c:
> Since this has been merged, is there still need of logging?

If you want to investigate the failure, you could add some logging and create a new PR that only contains the logging. Then I can repeat the build until it fails and we can investigate the logs.

abdulhannanali:
@marco-c Cool! I will do that, I am also interested in a test suite rewrite if you can initially guide me through it. Test suite is breaking often and maybe while rewriting test suite we can also refactor some of the code that's in the application.

PR #287, opened 2017-07-29
Add option to add Custom Expiration

abdulhannanali:
This PR let's the people add Custom Expiration to their vapid headers, but it seems expiration is a completed part and this is just the initial PR, the validation of the expiration can be more complex too.

The `validateExpiration` does not check the time for the minimum since minimums can be different in all points of time.

### Are there tests?

I have added a basic test suite of valid and invalid expiration examples, following the pattern in `test-vapid-helpers.js`, if  there are updates to be made, please himmeup with a comment.

### Timezone Differences

Date.now() returns time in UTC since UNIX Epoch as we all know. Are we guaranteed the push service is going to parse this in the UTC and not other timezone? The expiry parsing on the Push Service shouldn't be dependent on the timezone of the Server?

UPDATE: It's guaranteed to be [NumericDate](https://tools.ietf.org/html/rfc7519#page-6) which is in UTC and is the number of seconds since UNIX Epoch. 

### Updating expiry value everytime?

If a custom `expiration` value is not specified, it's going to update like every time we call the function, there's no benefit to it, since it gets updated by some time, even if seconds. Can we create a new expiration only once the one we currently generated is no longer valid?

### DIFF
--- a/src/vapid-helper.js
+++ b/src/vapid-helper.js
@@ -6,6 +6,12 @@ const asn1 = require('asn1.js');
 const jws = require('jws');
 const url = require('url');
 
+/**
+ * DEFAULT_EXPIRATION is set to seconds in 12 hours
+ */
+const DEFAULT_EXPIRATION_SECONDS = 12 * 60 * 60;
+const MAX_EXPIRATION_SECONDS = 24 * 60 * 60;
+
 const ECPrivateKeyASN = asn1.define('ECPrivateKey', function() {
   this.seq().obj(
     this.key('version').int(),
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
 }
 
 /**
+ * Calculates the total expiration by adding
+ * value of Date.now() with the seconds passed
+ *
+ * @param {Number} numSeconds Number of seconds to calculate expiration
+ * @return {Number} expiration in seconds
+ */
+function calculateExpiration(numSeconds) {
+  return Math.floor(Date.now() / 1000) + numSeconds;
+}
+
+/**
+ * Validates the Expiration based on the VAPID Spec
+ *
+ * `expiration` is not validated with a `minimum` value,
+ * since the time at validation is going to be different
+ * than at the time of creation of the custom expiration value
+ */
+function validateExpiration(expiration) {
+  if (!expiration) {
+    throw new Error('No expiration value provided in `expiration`');
+  }
+
+  expiration = Number.parseInt(expiration, 10);
+
+  if (isNaN(expiration) || typeof expiration !== 'number') {
+    throw new Error('Invalid expiration value passed, expiration must be `Number`');
+  }
+
+  // Roughly checks the time of expiration, since the max expiration can be ahead
+  // of the time than at the moment the expiration was generated
+  const maxExpiration = calculateExpiration(MAX_EXPIRATION_SECONDS);
+
+  if (expiration > maxExpiration) {
+    throw new Error('`expiration` is greater than maximum of 24 hours');
+  }
+
+  if (expiration < 0) {
+    throw new Error('`expiration` must be positive integer');
+  }
+}
+
+/**
  * This method takes the required VAPID parameters and returns the required
  * header to be added to a Web Push Protocol Request.
  * @param  {string} audience       This must be the origin of the push service.
@@ -123,11 +171,10 @@ function getVapidHeaders(audience, subject, publicKey, privateKey, expiration) {
   publicKey = urlBase64.decode(publicKey);
   privateKey = urlBase64.decode(privateKey);
 
-  const DEFAULT_EXPIRATION = Math.floor(Date.now() / 1000) + 43200;
-
   if (expiration) {
-    // TODO: Check if expiration is valid and use it in place of the hard coded
-    // expiration of 24hours.
+    validateExpiration(expiration);
+  } else {
+    expiration = calculateExpiration(DEFAULT_EXPIRATION_SECONDS);
   }
 
   const header = {
@@ -137,7 +184,7 @@ function getVapidHeaders(audience, subject, publicKey, privateKey, expiration) {
 
   const jwtPayload = {
     aud: audience,
-    exp: DEFAULT_EXPIRATION,
+    exp: expiration,
     sub: subject
   };
 
--- a/src/vapid-helper.js
+++ b/src/vapid-helper.js
@@ -205,5 +205,6 @@ module.exports = {
   getVapidHeaders: getVapidHeaders,
   validateSubject: validateSubject,
   validatePublicKey: validatePublicKey,
-  validatePrivateKey: validatePrivateKey
+  validatePrivateKey: validatePrivateKey,
+  validateExpiration: validateExpiration
 };
--- a/src/vapid-helper.js
+++ b/src/vapid-helper.js
@@ -10,6 +10,8 @@ const url = require('url');
  * DEFAULT_EXPIRATION is set to seconds in 12 hours
  */
 const DEFAULT_EXPIRATION_SECONDS = 12 * 60 * 60;
+
+// Maximum expiration is 24 hours according. (See VAPID spec)
 const MAX_EXPIRATION_SECONDS = 24 * 60 * 60;
 
 const ECPrivateKeyASN = asn1.define('ECPrivateKey', function() {
@@ -96,44 +98,40 @@ function validatePrivateKey(privateKey) {
 }
 
 /**
- * Calculates the total expiration by adding
- * value of Date.now() with the seconds passed
+ * Given the number of seconds calculates
+ * the expiration in the future by adding the passed `numSeconds`
+ * with the current seconds from Unix Epoch
  *
- * @param {Number} numSeconds Number of seconds to calculate expiration
- * @return {Number} expiration in seconds
+ * @param {Number} numSeconds Number of seconds to be added
+ * @return {Number} Future expiration in seconds
  */
-function calculateExpiration(numSeconds) {
-  return Math.floor(Date.now() / 1000) + numSeconds;
+function getFutureExpirationTimestamp(numSeconds) {
+  const futureExp = new Date();
+  futureExp.setSeconds(futureExp.getSeconds() + numSeconds);
+  return Math.floor(futureExp.getTime() / 1000);
 }
 
 /**
- * Validates the Expiration based on the VAPID Spec
+ * Validates the Expiration Header based on the VAPID Spec
+ * Throws error of type `Error` if the expiration is not validated
  *
- * `expiration` is not validated with a `minimum` value,
- * since the time at validation is going to be different
- * than at the time of creation of the custom expiration value
+ * @param {Number} expiration Expiration seconds from Epoch to be validated
  */
 function validateExpiration(expiration) {
-  if (!expiration) {
-    throw new Error('No expiration value provided in `expiration`');
+  if (!Number.isInteger(expiration)) {
+    throw new Error('`expiration` value must be a number');
   }
 
-  expiration = Number.parseInt(expiration, 10);
-
-  if (isNaN(expiration) || typeof expiration !== 'number') {
-    throw new Error('Invalid expiration value passed, expiration must be `Number`');
+  if (expiration < 0) {
+    throw new Error('`expiration` must be a positive integer');
   }
 
   // Roughly checks the time of expiration, since the max expiration can be ahead
   // of the time than at the moment the expiration was generated
-  const maxExpiration = calculateExpiration(MAX_EXPIRATION_SECONDS);
+  const maxExpiration = getFutureExpirationTimestamp(MAX_EXPIRATION_SECONDS);
 
-  if (expiration > maxExpiration) {
-    throw new Error('`expiration` is greater than maximum of 24 hours');
-  }
-
-  if (expiration < 0) {
-    throw new Error('`expiration` must be positive integer');
+  if (expiration >= maxExpiration) {
+    throw new Error('`expiration` value is greater than maximum of 24 hours');
   }
 }
 
@@ -174,7 +172,7 @@ function getVapidHeaders(audience, subject, publicKey, privateKey, expiration) {
   if (expiration) {
     validateExpiration(expiration);
   } else {
-    expiration = calculateExpiration(DEFAULT_EXPIRATION_SECONDS);
+    expiration = getFutureExpirationTimestamp(DEFAULT_EXPIRATION_SECONDS);
   }
 
   const header = {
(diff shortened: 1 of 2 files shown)


abdulhannanali:
@marco-c @gauntface  Would love if any one of you can take a look.

abdulhannanali:
### Why is there no check for the minimum expiration?

In [Section 4.1.4 of RFC JsonWebToken `exp` claim], there's a statement:
```
Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew.  Its value MUST be a number containing a NumericDate value.  Use of this claim is OPTIONAL.
```

There are Push Service Implementer here who can guide whether the Push Services provide this leeway or not. Otherwise, in order to be more, precise and strict. I guess we can add a simple comparison for the expiration with current date

```js
(expiration < Date.now() )
```

abdulhannanali:
Can we have a leeway as a result to compare the clock skew?

src/vapid-helper.js
```diff
@@ -6,6 +6,12 @@ const asn1 = require('asn1.js');
 const jws = require('jws');
 const url = require('url');
 
+/**
+ * DEFAULT_EXPIRATION is set to seconds in 12 hours
+ */
+const DEFAULT_EXPIRATION_SECONDS = 12 * 60 * 60;
+const MAX_EXPIRATION_SECONDS = 24 * 60 * 60;
```
marco-c: Why enforcing a maximum of one day?

src/vapid-helper.js
```diff
@@ -6,6 +6,12 @@ const asn1 = require('asn1.js');
 const jws = require('jws');
 const url = require('url');
 
+/**
+ * DEFAULT_EXPIRATION is set to seconds in 12 hours
+ */
+const DEFAULT_EXPIRATION_SECONDS = 12 * 60 * 60;
```
marco-c: Can we keep the default as one day?

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
 
 /**
+ * Calculates the total expiration by adding
+ * value of Date.now() with the seconds passed
+ *
+ * @param {Number} numSeconds Number of seconds to calculate expiration
+ * @return {Number} expiration in seconds
+ */
+function calculateExpiration(numSeconds) {
+  return Math.floor(Date.now() / 1000) + numSeconds;
```
marco-c: To be more precise, you could do something like:
```
let date = Date.now();
date.setSeconds(date.getSeconds() + numSeconds);
return date.getTime();
```

src/vapid-helper.js
marco-c: Right, the default was 12 hours, despite the comment :)

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
 }
 
 /**
+ * Calculates the total expiration by adding
+ * value of Date.now() with the seconds passed
+ *
+ * @param {Number} numSeconds Number of seconds to calculate expiration
+ * @return {Number} expiration in seconds
+ */
+function calculateExpiration(numSeconds) {
```
marco-c: Nit: I'd like a nicer name here, this is a function returning the current date + numSeconds, it's not stricly to calculate expiration.

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+ * Validates the Expiration based on the VAPID Spec
+ *
+ * `expiration` is not validated with a `minimum` value,
+ * since the time at validation is going to be different
+ * than at the time of creation of the custom expiration value
+ */
+function validateExpiration(expiration) {
+  if (!expiration) {
+    throw new Error('No expiration value provided in `expiration`');
+  }
```
marco-c: Nit: This check is redundant with the one in the caller.

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+  // Roughly checks the time of expiration, since the max expiration can be ahead
+  // of the time than at the moment the expiration was generated
+  const maxExpiration = calculateExpiration(MAX_EXPIRATION_SECONDS);
+
+  if (expiration > maxExpiration) {
+    throw new Error('`expiration` is greater than maximum of 24 hours');
+  }
+
+  if (expiration < 0) {
+    throw new Error('`expiration` must be positive integer');
```
marco-c: Nit: "`expiration` must be a positive integer

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+
+  if (isNaN(expiration) || typeof expiration !== 'number') {
+    throw new Error('Invalid expiration value passed, expiration must be `Number`');
+  }
+
+  // Roughly checks the time of expiration, since the max expiration can be ahead
+  // of the time than at the moment the expiration was generated
+  const maxExpiration = calculateExpiration(MAX_EXPIRATION_SECONDS);
+
+  if (expiration > maxExpiration) {
```
gauntface: This should be >=

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+    throw new Error('`expiration` is greater than maximum of 24 hours');
+  }
+
+  if (expiration < 0) {
```
gauntface: Nit: I'd move this before calculating and checking the max expiration.

test/test-vapid-helper.js
```diff
@@ -85,6 +85,18 @@ suite('Test Vapid Helpers', function() {
       },
       function() {
         vapidHelper.getVapidHeaders(VALID_AUDIENCE, { something: 'else' }, VALID_PUBLIC_KEY, VALID_PRIVATE_KEY);
+      },
+      function () {
+        vapidHelper.getVapidHeaders(VALID_AUDIENCE, VALID_SUBJECT_MAILTO, VALID_PUBLIC_KEY, VALID_PRIVATE_KEY, 'This is not valid expiration');
```
gauntface: Nit: could you change to `Not valid expiration: must be a number passed in as a header string, this is string with text`

test/test-vapid-helper.js
```diff
@@ -85,6 +85,18 @@ suite('Test Vapid Helpers', function() {
...
+      },
+      function () {
+        vapidHelper.getVapidHeaders(VALID_AUDIENCE, VALID_SUBJECT_MAILTO, VALID_PUBLIC_KEY, VALID_PRIVATE_KEY, { message: 'Expiration should be a number' });
```
gauntface: Nit: could you change to `Not valid expiration: must be a number passed in as a header string, this is an object`

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+}
+
+/**
+ * Validates the Expiration based on the VAPID Spec
+ *
+ * `expiration` is not validated with a `minimum` value,
+ * since the time at validation is going to be different
+ * than at the time of creation of the custom expiration value
+ */
+function validateExpiration(expiration) {
```
gauntface: expirationMSFromEpoch

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+  if (!expiration) {
+    throw new Error('No expiration value provided in `expiration`');
+  }
+
+  expiration = Number.parseInt(expiration, 10);
```
gauntface: Can you change this to:

```
const expirationEpochMS = Number.parseInt(expirationHeader, 10)
```

test/test-vapid-helper.js
```diff
@@ -109,6 +121,12 @@ suite('Test Vapid Helpers', function() {
       function() {
         return vapidHelper.getVapidHeaders(VALID_AUDIENCE, VALID_SUBJECT_URL,
           VALID_PUBLIC_KEY, VALID_PRIVATE_KEY, VALID_EXPIRATION);
+      },
+      function() {
+        return vapidHelper.getVapidHeaders(VALID_AUDIENCE, VALID_SUBJECT_URL, VALID_PUBLIC_KEY, VALID_PRIVATE_KEY, 0);
+      },
+      function () {
+        return vapidHelper.getVapidHeaders(VALID_AUDIENCE, VALID_SUBJECT_URL, VALID_PUBLIC_KEY, VALID_PRIVATE_KEY, Math.floor(Date.now() / 1000) + (5 * 60 * 60));
```
gauntface: Could you pull out this values into a variable and add a comment whether this is a valid expiration or not.

Could you also ensure this is a string when passed in.

src/vapid-helper.js
gauntface: We orginally had 24 hours but we hit an issue where some peoples servers would get into a bad state with web push services (i.e. I think one provided - probably FCM - would reject expirations over 24 hours in the future)

src/vapid-helper.js
gauntface: I think VAPID must have an expiration less that 24 hours, but link to spec would be good here.

src/vapid-helper.js
```diff
@@ -90,6 +96,48 @@ function validatePrivateKey(privateKey) {
...
+
+  if (isNaN(expiration) || typeof expiration !== 'number') {
+    throw new Error('Invalid expiration value passed, expiration must be `Number`');
```
gauntface: Should the expiration value be a string or a number in this function? This PR suggests a string, but I'm not sure why.

src/vapid-helper.js
abdulhannanali: Yup! There's a maximum limit of 24 hours for the expiration. I am still confused whether we should check for the minimum or maximums at all. But according to the standard this MAX Check is fine. 

```
   o  An "exp" (Expiry) claim MUST be included with the time after which
      the token expires.  This limits the time over which a token is
      valid.  An "exp" claim MUST NOT be more than 24 hours from the
      time of the request.
```

in [Page 4](https://tools.ietf.org/html/draft-ietf-webpush-vapid-03#page-4) of VAPID spec under the **Application Server Self-Identification** heading.

src/vapid-helper.js
abdulhannanali: @gauntface Yup! FCM should I looked into the code of autopush and it does that too. It's standard spec behavior.

src/vapid-helper.js
abdulhannanali: @gauntface Yup! This explicitly parses a string passed to it into a number, if it's not already a number. But since it's causing confusion, this can be removed, so the option only accepts number.

src/vapid-helper.js
marco-c: OK, it makes sense then!

src/vapid-helper.js
abdulhannanali: @marco-c Was making this change, just for the record the above code has a typo. It should be `new Date()` instead of `Date.now()` and since we don't seem to be changing the reference to date. I think we can assume that we can declare it as a const.

abdulhannanali: Thanks for the example. I did not know about this neat way to save myself from milliseconds to seconds and vice versa conversion.

src/vapid-helper.js
abdulhannanali: Thanks a lot for this! Yup! I can clearly now see how useless this check was.

src/vapid-helper.js
abdulhannanali: @marco-c We don't want the time according to the local timezone of the server, we need UTC time and that's why we should use UTC equivalent methods for everything. 

The `date.setSeconds` accepts a value in the range of 0 to 59 while the number of seconds we are specifying here are going to be much more than that. Just for the sake of example
```
12 * 60 * 60 = 43200
```

Similarly `date.getSeconds` provides a value in the range of 0 to 59. Confused a bit about it. 

Here `numSeconds` is the number of seconds till the expiration which might well be a value like this one `1501807169` time in seconds 24 hours at the moment I am typing. 

I think this confusion is happening because of the confusing name, as well as the confusing comment.

src/vapid-helper.js
abdulhannanali: Agreed

src/vapid-helper.js
abdulhannanali: Very good idea! Thanks

src/vapid-helper.js
abdulhannanali: @gauntface Sorry for this confusion, I have removed the assumption that it could be string at the first place, as It was confusing, and the Comment Signature for the getVapidHeaders also suggested that expiration must be `integer` so is it still needed?  Since it's no longer being parsed from string, we might not need two terms. However, I am still open to add `expirationFromSeconds` if that makes sense. Would love your opinion on it. Thanks :+1:

src/vapid-helper.js
abdulhannanali: Removed the assumption of String.

test/test-vapid-helper.js
abdulhannanali: This is contained within the validInputs, however in the latest commit I just added more comments and separated out this value into a variable

abdulhannanali:
Thanks for the comprehensive reviews @gauntface and @marco-c I got to learn a lot from it :smile: and appreciate them. also sorry for all these nits, I have done a few changes based on the code reviews and made some things more clearer.

src/vapid-helper.js
```diff
@@ -90,6 +98,44 @@ function validatePrivateKey(privateKey) {
...
+ * the expiration in the future by adding the passed `numSeconds`
+ * with the current seconds from Unix Epoch
+ *
+ * @param {Number} numSeconds Number of seconds to be added
+ * @return {Number} Future expiration in seconds
+ */
+function getFutureExpirationTimestamp(numSeconds) {
+  const futureExp = new Date();
+  futureExp.setSeconds(futureExp.getSeconds() + numSeconds);
+  return Math.floor(futureExp.getTime() / 1000);
```
marco-c: Nit: You can use `getUTCSeconds()` instead of `Math.floor(futureExp.getTime() / 1000)`.

abdulhannanali: @marco-c actually I should use setUTCSeconds and getUTCSeconds in the line 110 but It's still not clear to me if getUTCSeconds is what we need here, we need futureExpirationTimestamp which is calculated by adding the current seconds in UNIX epoch with the `numSeconds` 

getUTCSeconds is going to return a value between 0 and 59. Does that clear it out?

marco-c: Right, leave it as `Math.floor(futureExp.getTime() / 1000)`.

Line 110 should be fine, since you are adding seconds it doesn't matter if you are adding them in UTC or non-UTC, the number of seconds you add is going to be the same.

abdulhannanali: Yup! Good point! Sounds good! Thanks!

PR #296, opened 2017-10-17
Fixes #295

### LINKED ISSUE #295: Vapid tests intermittently fail with "Error: Vapid private key should be 32 bytes long when decoded."
We have seen this in multiple instances (e.g. in #286 and in #294). I'm attaching some logs before they expire.
[log1.txt](https://github.com/web-push-libs/web-push/files/1307961/202306396.txt)
[log2.txt](https://github.com/web-push-libs/web-push/files/1307966/log2.txt)
[log3.txt](https://github.com/web-push-libs/web-push/files/1307967/log3.txt)

gauntface:


marco-c:
Good find! Could you write a test for this, if it's not too much work? You could mock getPublicKey and getPrivateKey to return an invalid one and check that it gets padded.

gauntface:
Ask and you shall receive.

This was the best I could come up with in terms of stubbing / mocking.

marco-c:
> Ask and you shall receive.

> This was the best I could come up with in terms of stubbing / mocking.

Thanks, looks great!

PR #300, opened 2017-11-29
Accept any 20* response status as successful

### LINKED ISSUE #299: Accept 202 return code as a successful return code
We are only considering 201 as an expected result:
https://github.com/web-push-libs/web-push/blob/c998e20d875a97adde556604a097fd99800d1640/src/web-push-lib.js#L274

Actually, the push service could return 202 too.

marco-c:
Fixes #299.

CC @jrconlin


jrconlin:
👍

PR #309, opened 2018-01-25
Support aes128gcm

aliams:


marco-c:
Thanks @aliams.
There are still a couple of tests failing.
We should also add some additional tests to test this encoding, at least in test/testSelenium.js and possibly in test/testSendNotification.js too.
Last but not least, the documentation should be updated with the new option.

aliams:
I made some adjustments to the tests to make sure they can accommodate the new parameters. I got them mostly passing, but it seems that Ruby Node 6 and Stable and are failing on node --harmony ./test/helpers/download-test-browsers.js. Any ideas?

marco-c:
Not sure what's wrong with the two Mac tests, but it seems like they are failing while downloading the browsers for the Selenium tests. CC @gauntface

src/vapid-helper.js
```diff
@@ -210,10 +211,18 @@ function getVapidHeaders(audience, subject, publicKey, privateKey, expiration) {
     privateKey: toPEM(privateKey)
   });
 
-  return {
+  let vapidHeaders = {
```
marco-c: Let's do something like:
```
let vapidHeaders;
if (contentEncoding == 'aesgcm') {
  ...
} else if (contentEncoding == 'aes128gcm') {
  ...
} else {
  throw ...
}
```

src/web-push-lib.js
```diff
@@ -18,6 +18,11 @@ function WebPushLib() {
 
 }
 
+WebPushLib.supportedContentEncodings = {
```
marco-c: I wish this was available in vapid-helper.js too, so that we wouldn't need to hardcode the strings.

src/web-push-lib.js
```diff
@@ -201,6 +223,10 @@ WebPushLib.prototype.generateRequestDetails =
         requestDetails.headers.Authorization = 'key=' + currentGCMAPIKey;
       }
     } else if (currentVapidDetails) {
+      if (contentEncoding === WebPushLib.supportedContentEncodings.AES_128_GCM
+          && subscription.endpoint.indexOf('https://fcm.googleapis.com') === 0) {
+        subscription.endpoint = subscription.endpoint.replace('fcm/send', 'wp');
```
marco-c: @gauntface is this reliable?

src/web-push-lib.js
aliams: Should I just require the web push lib in vapid-helper.js?

test/test-encryption-helper.js
```diff
@@ -15,26 +15,25 @@ suite('Test Encryption Helpers', function() {
...
 
   test('encrypt/decrypt string', function() {
-    assert(encryptDecrypt('hello').equals(new Buffer('hello')));
+    assert(encryptDecrypt('hello', webPush.supportedContentEncodings.AES_GCM).equals(new Buffer('hello')));
   });
 
   test('encrypt/decrypt buffer', function() {
-    assert(encryptDecrypt(new Buffer('hello')).equals(new Buffer('hello')));
+    assert(encryptDecrypt(new Buffer('hello'), webPush.supportedContentEncodings.AES_GCM).equals(new Buffer('hello')));
   });
```
marco-c: Can you add tests with AES128GCM?

src/web-push-lib.js
aliams: For now it is. We need to do this until FCM actually supports it on the normal URL.

src/vapid-helper.js
aliams: Sounds good to me.

test/testSelenium.js
```diff
@@ -157,15 +157,17 @@ function runTest(browser, options) {
...
         if (!subscription.keys) {
           throw new Error('Require subscription.keys not found.');
         }
 
         promise = webPush.sendNotification(subscription, pushPayload, {
-          vapidDetails: vapid
+          vapidDetails: vapid,
+          contentEncoding: 'aesgcm'
         });
       }
```
marco-c: We should add tests with aes128gcm too

test/testSendNotification.js
```diff
@@ -147,14 +147,12 @@ suite('sendNotification', function() {
...
-
       const decrypted = ece.decrypt(requestBody, {
-        keyid: 'webpushKey',
+        version: 'aesgcm',
+        privateKey: userCurve,
         dh: appServerPublicKey,
         salt: salt,
-        authSecret: urlBase64.encode(userAuth),
-        padSize: 2
+        authSecret: urlBase64.encode(userAuth)
```
marco-c: And we should add tests with aes128gcm here too (but testSelenium is the main one where we need it).

src/web-push-lib.js
marco-c: I don't know if that would work, as web-push-lib is importing vapid-helper, so this would create an infinite loop.

Maybe we can put this variable into a separate web-push-constants.js module (even though it would be a really small module 😄 )

aliams:
I expanded the tests and addressed the feedback you had and updated the readme. Please take a look and let me know.

I think the tests will require more work to ensure everything is being tested correctly. For the time being, I've just doubled the aesgcm tests so that we can cover aes128gcm as well.

marco-c:
Thanks!

The only failing tests are unrelated and tracked in #312.

PR #313, opened 2018-01-29
Add aes128gcm support for CLI

aliams:
I already added the support for CLI in the previous PR, but I didn't update the docs and test. This PR updates the docs and the test.


marco-c:
Thanks!

PR #319, opened 2018-03-16
Add proxy option

### LINKED ISSUE #280: ETIMEDOUT in sendNotification.Can it add a proxy option?

andolini:
Fixes #280.


src/web-push-lib.js
```diff
@@ -285,7 +287,11 @@ WebPushLib.prototype.sendNotification =
 
       httpsOptions.headers = requestDetails.headers;
       httpsOptions.method = requestDetails.method;
-
+      
+      if('proxy' in options && typeof options.proxy === 'string') {
+          httpsOptions.agent = new HttpsProxyAgent(options.proxy);
```
gauntface: Could you separate the check for the type of `options.proxy` and print a console warning if it's set but not used because it's not a string.

I'd also be happy with throwing an error.

gauntface:
Could you also add a test (along the lines of https://github.com/web-push-libs/web-push/blob/61dddeb4b55eaa3b70072b85a47cea1442bf0b80/test/test-generate-request-details.js#L228)

test/test-generate-request-details.js
```diff
@@ -233,7 +233,8 @@ suite('Test Generate Request Details', function() {
       headers: {
         'Topic': 'topic',
         'Urgency': 'urgency'
-      }
+      },
+      'proxy': 'proxy'
```
marco-c: A separate test should be added for the proxy option, this test is for the **Extra headers** parameter.

marco-c:
It would be nice to have a test in `test/testSendNotification.js` too, but we can do it in a follow-up.

marco-c:
The Travis build is failing with:
```
Error: Cannot find module 'ms'
    at Function.Module._resolveFilename (module.js:543:15)
    at Function.Module._load (module.js:470:25)
    at Module.require (module.js:593:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/travis/build/web-push-libs/web-push/node_modules/debug/src/debug.js:14:20)
    at Module._compile (module.js:649:30)
    at Object.Module._extensions..js (module.js:660:10)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:501:12)
    at Function.Module._load (module.js:493:3)
```

package-lock.json
```diff
@@ -4,10 +4,25 @@
   "lockfileVersion": 1,
```
marco-c: This file has a lot of changes that might be unnecessary.

marco-c:
Merged manually in f099ec8ff97e86fb6778ece04bd27a36ef93655e. Thanks!

PR #393, opened 2018-10-31
Remove useless escape

FluorescentHallucinogen:
Escaping non-special characters in regular expressions doesn’t have any effect.


FluorescentHallucinogen:
@marco-c PTAL.

PR #395, opened 2018-12-10
fcm/send/ endpoint supports AES_128_GCM

pinarx:
Remove the FCM endpoint modification for AES_128_GCM content encoding

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -230,10 +230,6 @@ WebPushLib.prototype.generateRequestDetails =
         requestDetails.headers.Authorization = 'key=' + currentGCMAPIKey;
       }
     } else if (currentVapidDetails) {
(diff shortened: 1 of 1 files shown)


marco-c:
@pinarx do you have a reference that shows that this is now supported on the "normal" URL?

pinarx:
You can confirm it using this demo and by simply replacing the endpoint to point to /fcm/send rather than /wp/.

MartijnDwars:
Does this mean Google made a change such that this replacement is no longer necessary? The Selenium tests only check with newly created subscriptions; what if we have an old endpoint that still contains `/fcm/send`?

pinarx:
The replacement is no longer necessary and the /fcm/send endpoint now supports the new VAPID format. Old tokens should continue to work.

PR #415, opened 2019-02-05
FCM endpoint support for GCM/FCM projects

sarpdoruk:
While using a GCM or FCM project, sending legacy HTTP request to`https://fcm.googleapis.com/fcm/send` endpoint won't work. FCM returns `Unauthorized` response. In this very case GCM endpoint works fine. An endpoint check is added for FCM endpoint support.


src/web-push-lib.js
```diff
@@ -220,16 +220,9 @@ WebPushLib.prototype.generateRequestDetails =
     }
 
     const isGCM = subscription.endpoint.indexOf('https://android.googleapis.com/gcm/send') === 0;
+    const isFCM = subscription.endpoint.indexOf('https://fcm.googleapis.com/fcm/send') === 0;
     // VAPID isn't supported by GCM hence the if, else if.
-    if (isGCM) {
```
marco-c: VAPID isn't supported by GCM, so we should keep this if first

src/web-push-lib.js
```diff
@@ -252,6 +245,14 @@ WebPushLib.prototype.generateRequestDetails =
           requestDetails.headers['Crypto-Key'] = vapidHeaders['Crypto-Key'];
         }
       }
+    } else if (isGCM || isFCM) {
```
marco-c: Are you sure this works with FCM?

sarpdoruk: Yes it does work with FCM, that is why I'm trying to add as a supported webpush endpoint.

src/web-push-lib.js
sarpdoruk: There is a tradeoff here; if you are using VAPID, then you shouldn't be using GCM endpoint. 

If you are using FCM endpoint to send pushes but without VAPID then this library becomes useless because of the `isGCM` check at the top since FCM is also supported for regular webpushes.

The modification I made allows users to send a regular webpush, w/o VAPID, to both GCM and FCM endpoints which work fine. It does not affect pushed using VAPID.

marco-c: Wouldn't something like:
```JavaScript
if (isGCM) {
  // Use GCM API key
  ...
  // Use VAPID
  ...
} else if (isFCM and currentGCMAPIKey) {
  // Use GCM API key
  ...
}
```
work for you?

sarpdoruk: Yes it would but;
```js
if (isGCM) {
  // The Push logic in this clause is exactly the SAME
} else if (isFCM and currentGCMAPIKey) {
  // with the ONE here
}
```

That is why I'm trying to have;

```js
if (isGCM || isFCM) {
  ...
}
```

marco-c: Yes, I see your point, but I feel more comfortable with this change if we don't change the pre-existing behavior at all, even if the code is almost duplicated.

sarpdoruk: Oh OK, that is your call. Would you like me to commit these changes you wanted ?

marco-c: Yes, thanks!

marco-c: @sarpdoruk will you update the PR?

sarpdoruk: Today, I will.

PR #456, opened 2019-05-13
Allow falsy values for vapidDetails

### LINKED ISSUE #445: Migrated GCM endpoints don't work
**NOTE**: Please test in a least two browsers (i.e. Chrome and Firefox). This
helps with diagnosing problems quicker.

# Setup

> Please provide the following details, the more info you can provide the
> better.

*Operating System*: Linux
*Node Version*: v10.13.0
*web-push Version*: `master`

> Please select any browsers that you are experiencing problems with:

- [x] Chrome
- [ ] Firefox
- [ ] Opera for Android
- [ ] Samsung Internet Browser
- [ ] Other

Server side problem while sending pushes for migrated endpoints.

# Problem
I had a mix of old GCM endpoints and newer endpoints using vapid.

Google deprecated GCM on April 10th 2019, and according to this [migration guide](https://firebase.googleblog.com/2018/04/time-to-upgrade-from-gcm-to-fcm.html) suggested to edit all server endpoints to FCM ones (Step 4).

`web-push` library added support for FCM in #415 and it's priority is below VAPID. But this leads to the old migrated endpoints being sent with Vapid headers and failing. If it would be above Vapid then the newer endpoints registered through Vapid will fail.

# Expected

I think there should be an option to select the method to send the push with. Or allow `false` value in options for the `sendNotification` method so that I can set `vapidDetails` to `false` for the old migrated GCM(now FCM) endpoints.

# Features Used

- [x] VAPID Support
- [x] GCM API Key
- [x] Sending with Payload

rohit-smpx:
Closes #445 

I added some tests for multiple FCM cases, but I had to fix the vapid validation for that. Please review those changes once.

Also should I separate the tests for FCM from current GCM tests?

### DIFF
--- a/test/testSendNotification.js
+++ b/test/testSendNotification.js
@@ -127,6 +127,8 @@ suite('sendNotification', function() {
 
   function validateRequest(request) {
     const options = request.requestOptions;
+    const contentEncoding = (options.extraOptions && options.extraOptions.contentEncoding)
+      || WebPushConstants.supportedContentEncodings.AES_GCM;
     const isGCM = options.subscription.endpoint
       .indexOf('https://android.googleapis.com/gcm') === 0;
 
@@ -152,10 +154,10 @@ suite('sendNotification', function() {
         return key.indexOf('dh=') === 0;
       }).substring('dh='.length);
 
-      assert.equal(requestDetails.headers['content-encoding'], options.extraOptions.contentEncoding, 'Check Content-Encoding header');
+      assert.equal(requestDetails.headers['content-encoding'], contentEncoding, 'Check Content-Encoding header');
 
       const decrypted = ece.decrypt(requestBody, {
-        version: options.extraOptions.contentEncoding,
+        version: contentEncoding,
         privateKey: userCurve,
         dh: appServerPublicKey,
         salt: salt,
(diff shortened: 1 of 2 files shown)


test/testSendNotification.js
```diff
@@ -166,29 +170,40 @@ suite('sendNotification', function() {
...
-      const decoded = jws.decode(jwt);
-      assert.equal(decoded.header.typ, 'JWT');
-      assert.equal(decoded.header.alg, 'ES256');
-      assert.equal(decoded.payload.aud, 'https://127.0.0.1');
-      assert(decoded.payload.exp > Date.now() / 1000);
-      assert.equal(decoded.payload.sub, 'mailto:mozilla@example.org');
+       const decoded = jws.decode(jwt);
+       assert.equal(decoded.header.typ, 'JWT');
+       assert.equal(decoded.header.alg, 'ES256');
+       assert.equal(options.subscription.endpoint.indexOf(decoded.payload.aud) === 0, true);
```
marco-c: Nit: use startsWith instead

test/testSendNotification.js
```diff
@@ -166,29 +170,40 @@ suite('sendNotification', function() {
...
+        assert.equal(authorizationHeader.indexOf('WebPush '), 0, 'Check VAPID Authorization header');
+        jwt = authorizationHeader.substring('WebPush '.length);
+      } else if (contentEncoding === WebPushConstants.supportedContentEncodings.AES_128_GCM) {
+        const authorizationHeader = requestDetails.headers.authorization;
+        assert.equal(authorizationHeader.indexOf('vapid t='), 0, 'Check VAPID Authorization header');
+        [jwt, vapidKey] = authorizationHeader.substring('vapid t='.length).split(', k=');
+      }
 
-      assert(appServerVapidPublicKey.equals(vapidKeys.publicKey));
+      // const appServerVapidPublicKey = urlBase64.decode(vapidKey);
```
marco-c: Nit: this is not needed anymore

marco-c:
> Also should I separate the tests for FCM from current GCM tests?

No, it's OK like this.

PR #458, opened 2019-05-14
Allow proxy options object as a proxy parameter

### LINKED ISSUE #450: Proxy accepts only url string
# Setup

*Operating System*: Windows
*Node Version*: 8.12
*web-push Version*: 3.3.3

# Problem

Proxy option accepts only strings (web-push-lib.js:170) but not [HttpsProxyAgent options](https://github.com/TooTallNate/node-https-proxy-agent#new-httpsproxyagentobject-options)

# Expected

I expect it to accept either options object as a proxy parameter:
```javascript
const urlParse = require('url').parse;

const proxyOptions = urlParse('http://myproxy');
proxyOptions.rejectUnauthorized = false;

webPush.sendNotification(subscription, payload, {
  proxy: proxyOptions,
});
```

or HttpsProxyAgent as an agent parameter

```javascript
const HttpsProxyAgent = require('https-proxy-agent');
const urlParse = require('url').parse;

const proxyOptions = urlParse('http://myproxy');
const agent = new HttpsProxyAgent(proxyOptions);

webPush.sendNotification(subscription, payload, {
  agent: agent,
});
```

tim-kuteev:
Closes #450

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ const options = {
     '< header name >': '< header value >'
   },
   contentEncoding: '< Encoding type, e.g.: aesgcm or aes128gcm >',
-  proxy: '< proxy server address >'
+  proxy: '< proxy server options >'
 }
 
 webpush.sendNotification(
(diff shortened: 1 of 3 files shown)


marco-c:
Looks good to me, but could you update the docs too to mention that proxy could either be a string defining a server address or an object in the same format as the argument of the HttpsProxyAgent constructor?

Issue #284, opened 2017-06-23
TTL doesn't seem to be honored


sathio:
# Setup

*Operating System*: Linux
*Node Version*: v6.2.2
*web-push Version*: ^3.2.2

# Browser
- [x] Chrome

# Problem

TTL is not honored, if I set 7200 seconds it is still shown the next morning when I fire up the browser

# Expected

Notifications should expire after TTL seconds

# Features Used

- [x] VAPID Support
- [x] GCM API Key
- [x] Sending with Payload

# Example / Reproduce Case
    const payload = new Buffer('{"title": "'+urldecode(title)+'", "body": "'+urldecode(body)+'", "icon": "'+icon+'", "href": "'+href+'", "tag": "'+tag+'"}', 'utf8');

    const options = { TTL: 7200 }

    webpush.sendNotification(subscription, payload, options).then(function(response) {
      console.log(JSON.stringify(response));
    });

marco-c:
Which version of Chrome?

sathio:
Version 58.0.3029.110 (64-bit)

gauntface:
could you please file this on [crbug.com](https://bugs.chromium.org/p/chromium/issues/list)

marco-c:
@sathio did you file a bug on crbug.com? Can you link it here?

sathio:
Hi, I will file an issue today, but I'm not yet convinced this is on chrome/chromium side. 
Isn't there a way to understand what headers are sent exactly to the FCM server?

marco-c:
>  Isn't there a way to understand what headers are sent exactly to the FCM server?

You can add some logging here if you want: https://github.com/web-push-libs/web-push/blob/master/src/web-push-lib.js#L265.
We have tests that check the TTL header is present though, so it shouldn't be a bug on our side.

sathio:
Here's a detailed response, HEADERS are from the web-push-lib.js file as you've suggested.

    [
      {
        "message":"sent",
        "title":"'title'",
        "body":"'body'",
        "icon":"'icon'",
        "href":"'href'",
        "tag":"'tag'",
        "res":[
        "HEADERS { hostname: 'android.googleapis.com',",
         " port: null,",
          " path: '/gcm/send/c7qnskTew5w:[redacted]',",
          " headers:",
          " { TTL: 7200,",
          " 'Content-Length': 440,",
          " 'Content-Type': 'application/octet-stream',",
          " 'Content-Encoding': 'aesgcm',",
[...]

PR #492, opened 2019-08-09
TTL accepts non-negative integers including ZERO

### LINKED ISSUE #491: Zero TTL Not Working
According to the spec [here](https://tools.ietf.org/html/draft-ietf-webpush-protocol-12#section-5.2), a zero TTL means that the message is not stored by FCM and only delivered to endpoints that are online at that moment.

In my testing using the node web-push library, when I am setting TTL to ZERO in the options object as follows :
`const options = {
        gcmAPIKey: <api_key>,
        vapidDetails: {
            subject: 'mailto:example@web-push-node.org',
            publicKey: <public_key>,
            privateKey: <private_key>
        },
        TTL: 0
}`

But receiving the web-push even after keeping the device offline for half hour.

Tested for both latest Chrome and Firefox.

Please help!

renegadeHomie:
Fixes #491 
#284 

The if condition to check if options.TTL is present doesn't honour ZERO. This commit makes TTL accepts non-negative integers including ZERO.


src/web-push-lib.js
```diff
@@ -155,7 +155,7 @@ WebPushLib.prototype.generateRequestDetails = function(subscription, payload, op
         currentVapidDetails = options.vapidDetails;
       }
 
-      if (options.TTL) {
+      if (Number.isInteger(options.TTL) && options.TTL > -1) {
```
marco-c:
```suggestion
      if (options.TTL !== undefined) {
```

And then:
```
timeToLive = Number(options.TTL)
if (timeToLive < 0) {
  throw new Error("TTL should be a number and should be at least 0");
}
```

renegadeHomie: Sure, will add the suggested changes.

Before going ahead, I want to point out the implications this change will have on the Impression stats for people using the library.

So, currently whoever is sending TTL=0, the messages are actually getting delivered for a timespan for 4 weeks as that is the default.

**With this change, the stats being recorded by such end-users for their bulk campaigns sent using this library will drastically drop.**

Would want to hear your thoughts on the same.

src/web-push-lib.js
```diff
@@ -190,6 +193,7 @@ WebPushLib.prototype.generateRequestDetails = function(subscription, payload, op
       }
     }
 
+
```
marco-c: Nit: Please remove this unrelated newline

src/web-push-lib.js
marco-c: I think it's an acceptable drawback, we need to honor the choice of the users.

renegadeHomie: Agree, this aligns the library with the VAPID [spec](https://tools.ietf.org/html/draft-ietf-webpush-protocol-12#section-5.2)

renegadeHomie:
Is there anything else I should add i.e. update the README or something?

src/web-push-lib.js
```diff
@@ -155,8 +155,11 @@ WebPushLib.prototype.generateRequestDetails = function(subscription, payload, op
         currentVapidDetails = options.vapidDetails;
       }
 
-      if (options.TTL) {
-        timeToLive = options.TTL;
+      if (options.TTL !== undefined) {
+        timeToLive = Number(options.TTL);
+        if (timeToLive < 0) {
+          throw new Error("TTL should be a number and should be at least 0");
```
marco-c: Just one needed change, for our eslint rules these should be single quotes.

```suggestion
          throw new Error('TTL should be a number and should be at least 0');
```

renegadeHomie:
@marco-c Says branch is outdated. It looks good otherwise. Please merge the PR.

PR #502, opened 2019-09-10
Upload coverage to coveralls

### LINKED ISSUE #311: Setup code coverage collection

collierrgbsitisfise:
Fixes #311

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "devDependencies": {
     "chalk": "2.4.2",
     "chromedriver": "76.0.1",
-    "coveralls": "^3.0.6",
+    "coveralls": "3.0.6",
     "del": "5.1.0",
     "eslint": "6.3.0",
     "eslint-config-airbnb": "18.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -39,6 +40,7 @@
   "devDependencies": {
     "chalk": "2.4.2",
     "chromedriver": "76.0.1",
+    "coveralls": "^3.0.6",
```
marco-c: Could you fix the version, like we do for the other devDependencies?

package.json
```diff
@@ -47,6 +49,7 @@
     "istanbul": "0.4.5",
     "mkdirp": "0.5.1",
     "mocha": "6.2.0",
+    "mocha-lcov-reporter": "^1.3.0",
```
marco-c: Same here

package.json
collierrgbsitisfise: yes, sure

package.json
collierrgbsitisfise: yes, sure

collierrgbsitisfise:
@marco-c , hi just did an fix according CR, just curious - what is a reason lock versions of 
 devDep(only) ?

marco-c:
> @marco-c , hi just did an fix according CR, just curious - what is a reason to lock versions of
devDep(only) ?

I generally prefer locking dependencies, to make things reproducible. Since this is a library, the normal dependencies can't be locked (or all the projects depending on the library will inherit the locked dependencies), but we can lock the devDependencies at least.

marco-c:
Thanks!

PR #534, opened 2019-12-01
Require https-proxy-agent only when actually needed

xPaw:
`https-proxy-agent` patches core Node methods in `https` for whatever reason, and that breaks unrelated code. In our app we don't use proxy, and the mere fact that `web-push` requires the proxy package, it breaks other code.

See https://github.com/TooTallNate/node-agent-base/issues/35 and https://github.com/sindresorhus/got/issues/951

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -325,7 +325,7 @@ WebPushLib.prototype.sendNotification = function(subscription, payload, options)
       }
 
       if (requestDetails.proxy) {
-        const HttpsProxyAgent = require('https-proxy-agent');
+        const HttpsProxyAgent = require('https-proxy-agent'); // eslint-disable-line global-require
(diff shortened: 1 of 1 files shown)


marco-c:
It'd be nice to have a test for the proxy option (currently we don't have any), to ensure it doesn't break.

xPaw:
I don't have the motivation to look into how to achieve that, sorry.

I'm not sure `https-proxy-agent` is a good package to begin with, when it override core node methods.

xPaw:
Thanks. Will you be able to push a new release out with this?

marco-c:
> Thanks. Will you be able to push a new release out with this?

Yes, I'll do it shortly.

marco-c:
Ah, you'll need to add `// eslint-disable-line global-require` to avoid the eslint error.

marco-c:
I've released v3.4.2.

PR #549, opened 2020-01-17
remove useless seconds subscription.keys validation

gavrashenko:
There are two checks for the same subscription object. I added one extra check for the first condition and remove second one.


gavrashenko:
Hello everyone. Why 'Travis CI' is still throws errors is all pr's ? Could anyone fix it ?

marco-c:
There is a temporary issue in Travis CI, I think they need to update their APT keys for the repository containing Chrome.

marco-c:
Thanks!

PR #569, opened 2020-04-03
Optional timeout option for send notification requests

carpasse:


andyghiuta:
Should we expect a new version anytime soon with this change?

marco-c:
I've released 3.4.4.

PR #576, opened 2020-04-17
skip redundant code code, correct readme

4r7d3c0:
There's some areas in code which are redundant. Plus readme indicates that `aesgcm` is the default but in fact it's `aes128gcm`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ request only. This overrides any API key set via `setGCMAPIKey()`.
 - **TTL** is a value in seconds that describes how long a push message is
 retained by the push service (by default, four weeks).
 - **headers** is an object with all the extra headers you want to add to the request.
-- **contentEncoding** is the type of push encoding to use (e.g. 'aesgcm128gcm', by default, or 'aesgcm').
+- **contentEncoding** is the type of push encoding to use (e.g. 'aes128gcm', by default, or 'aesgcm').
 - **proxy** is the [HttpsProxyAgent's constructor argument](https://github.com/TooTallNate/node-https-proxy-agent#new-httpsproxyagentobject-options)
 that may either be a string URI of the proxy server (eg. http://< hostname >:< port >)
 or an "options" object with more specific properties.
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ request only. This overrides any API key set via `setGCMAPIKey()`.
 - **TTL** is a value in seconds that describes how long a push message is
 retained by the push service (by default, four weeks).
 - **headers** is an object with all the extra headers you want to add to the request.
-- **contentEncoding** is the type of push encoding to use (e.g. 'aesgcm', by default, or 'aes128gcm').
+- **contentEncoding** is the type of push encoding to use (e.g. 'aesgcm128gcm', by default, or 'aesgcm').
 - **proxy** is the [HttpsProxyAgent's constructor argument](https://github.com/TooTallNate/node-https-proxy-agent#new-httpsproxyagentobject-options)
 that may either be a string URI of the proxy server (eg. http://< hostname >:< port >)
 or an "options" object with more specific properties.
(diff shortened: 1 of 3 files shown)


src/vapid-helper.js
```diff
@@ -115,6 +115,7 @@ function validatePrivateKey(privateKey) {
   if (privateKey.length !== 32) {
     throw new Error('Vapid private key should be 32 bytes long when decoded.');
   }
+  return privateKey;
```
marco-c, line 118: I prefer to keep validatePrivateKey as a function without return

README.md
```diff
@@ -184,7 +184,7 @@ request only. This overrides any API key set via `setGCMAPIKey()`.
 - **TTL** is a value in seconds that describes how long a push message is
 retained by the push service (by default, four weeks).
 - **headers** is an object with all the extra headers you want to add to the request.
-- **contentEncoding** is the type of push encoding to use (e.g. 'aesgcm', by default, or 'aes128gcm').
+- **contentEncoding** is the type of push encoding to use (e.g. 'aesgcm128gcm', by default, or 'aesgcm').
```
marco-c, line 187:
```suggestion
- **contentEncoding** is the type of push encoding to use (e.g. 'aes128gcm', by default, or 'aesgcm').
```

4r7d3c0: oh ok 🏴󠁧󠁢󠁷󠁬󠁳󠁿 do you know by chance what changes I would have to make to the encryption snippet from https://developers.google.com/web/updates/2016/03/web-push-encryption to upgrade it from `aesgcm` to `aes128gcm`? i'm working on something else this is not for the PR. currently the code is like

```
/**
 *
 * @param {string} p256dh The client's public key.
 * @param {string} auth The client's private key.
 * @param {string|!Buffer} data Some data to send as payload.
 */
export default function encryption(p256dh, auth, data) {
  const clientPublicKey = Buffer.from(p256dh, 'base64')
  const clientAuthSecret = Buffer.from(auth, 'base64')

  const salt = randomBytes(16)

  // Node has ECDH built-in to the standard crypto library. For some languages
  // you may need to use a third-party library.
  const serverECDH = createECDH('prime256v1')
  const serverPublicKey = serverECDH.generateKeys()
  const sharedSecret = serverECDH.computeSecret(clientPublicKey)

  const authInfo = Buffer.from('Content-Encoding: auth\0')
  const prk = hkdf(clientAuthSecret, sharedSecret, authInfo, 32)

  // Derive the Content Encryption Key
  const contentEncryptionKeyInfo = createInfo('aesgcm', clientPublicKey, serverPublicKey)
  const contentEncryptionKey = hkdf(salt, prk, contentEncryptionKeyInfo, 16)

  // Derive the Nonce
  const nonceInfo = createInfo('nonce', clientPublicKey, serverPublicKey)
  const nonce = hkdf(salt, prk, nonceInfo, 12)

  const paddingLength = 0
  const padding = Buffer.alloc(2 + paddingLength)
  // The buffer must be only zeroes, except the length
  padding.fill(0)
  padding.writeUInt16BE(paddingLength, 0)

  // Create a buffer from our data, in this case a UTF-8 encoded string
  const plaintext = Buffer.from(data)
  const cipher = createCipheriv('id-aes128-GCM', contentEncryptionKey, nonce)

  const result = cipher.update(Buffer.concat([padding, plaintext], padding.length + plaintext.length))
  cipher.final()

  // Append the auth tag to the result - https://nodejs.org/api/crypto.html#crypto_cipher_getauthtag
  return {
    cipherText: Buffer.concat([result, cipher.getAuthTag()]),
    serverPublicKey,
    salt,
  }
}
```

4r7d3c0: actually i'll have a look in https://github.com/web-push-libs/encrypted-content-encoding/blob/master/nodejs/ece.js

marco-c:
Thanks for the contribution!
For any potential future contribution, I'd prefer if you split the changes in multiple PRs.

Issue #585, opened 2020-06-05
Fix timeout option and add socket timeout


carpasse:
**NOTE**: Please test in a least two browsers (i.e. Chrome and Firefox). This
helps with diagnosing problems quicker.

# Setup
Operating System: All
Node Version: 12+
web-push Version: ^3.4.4

# Problem

> Current timeout option as it is only raises a timeout event in the reques and does not abort the request. 
> Also missing a socketTimeout options to abort abort slow connections

# Expected

> There request should timeout if it does not complete before the specified timout.
> We should be able to set a request socket timeout to abort requests on slow connections

# Features Used

- [ ] VAPID Support
- [ ] GCM API Key
- [x] Sending with Payload

# Example / Reproduce Case

> Connect to an existing host but to a port that is blocked by the firewall that simply drops TCP SYN packets. For example, www.google.com:81

# Other

> This issue fixes what issue https://github.com/web-push-libs/web-push/issues/568 intended to fix. I am preparing a PR for it.

PR #586, opened 2020-06-05
fix timeout option and add socket timeout option

carpasse:
fixes https://github.com/web-push-libs/web-push/issues/585


README.md
```diff
@@ -180,7 +180,8 @@ any of the following values defined, although none of them are required.
...
 *privateKey* values defined. These values should follow the [VAPID Spec](https://tools.ietf.org/html/draft-thomson-webpush-vapid).
-- **timeout** is a value in milliseconds that specifies how long the library must wait for a response from the push service before timing out (by default undefined).
+- **timeout** is a value in milliseconds that specifies how long the library must wait for a response from the push service before timing out. On timeout, the request will be destroyed and the promise will be rejected with a meaningful error (by default undefined).
+- **socketTimeout** is a value in milliseconds that specifies the request's socket timeout. On timeout, The request will be destroyed and the promise will be rejected with a meaningful error. It's a common misconception that a socket timeout is the timeout to receive the full response. So if you have a socket timeout of 1 second, and a response comprised of 3 TCP packets, where each response packet takes 0.9 seconds to arrive, for a total response time of 2.7 seconds, then there will be no timeout. Once a socket 'timeout' triggers the request will be aborted by the library (by default undefined).
```
marco-c, line 184:
```suggestion
- **socketTimeout** is a value in milliseconds that specifies the request's socket timeout. On timeout, the request will be destroyed and the promise will be rejected with a meaningful error. It's a common misconception that a socket timeout is the timeout to receive the full response. So if you have a socket timeout of 1 second, and a response comprised of 3 TCP packets, where each response packet takes 0.9 seconds to arrive, for a total response time of 2.7 seconds, then there will be no timeout. Once a socket 'timeout' triggers the request will be aborted by the library (by default undefined).
```

src/web-push-lib.js
```diff
@@ -197,6 +199,10 @@ WebPushLib.prototype.generateRequestDetails = function(subscription, payload, op
       if (typeof options.timeout === 'number') {
         timeout = options.timeout;
       }
+
+      if (typeof options.socketTimeout === 'number') {
+        socketTimeout = options.socketT;
```
marco-c, line 204:
```suggestion
        socketTimeout = options.socketTimeout;
```

marco-c:
What is the use case for having two separate timeouts? I'd just keep one (the one you've renamed to socketTimeout)

README.md
carpasse: updated the text with your sugestion

src/web-push-lib.js
carpasse: 👍 cheers

carpasse:
@marco-c  A socket timeout will allow you to fail faster without having to wait for the timeout to trigger if the connection gets dropped at start or in between the push request.

For example.
Having a configuration of ***timeout*** **_15000_** and ***socketTimeout*** **_5000_** 
The request will aborted after 5 seconds if the push service drops the connection messages or if a socket waited for more than 5 seconds to get a packet. In which case there is a high change the connection was droped and there is no need to wait 15 seconds.

You can simulate this by trying to push a notification to a subscription with the wrong port. 

To test this, I tried to send a notification my demo subscription changing the port

_https://fcm.googleapis.com**:81**/fcm/send/dy8ksxFX658:APA91bFB1rj8OAXQ2QZXy7kTt-tQ1iDj6ZtJxfCs6Drvgb8MY6TlbWCG-tJG9yVY0Sk9dLkcaM3NDi5p7fQjy5Z0jFh21-8kI81cSfc2BCTbX8o7CJfluvYSjsNbEF4CWju8Gd0gJ5BC_

Google push service ignores the request done to wrong ports. The request failed after only 5 seconds instead of 15 seconds with a socket timeout error

```
err: {
stack: 'Error: Socket timeout\n' +
  '    at ClientRequest.<anonymous> (/Users/carlos/gh/carpasse/web-push-api/node_modules/web-push/src/web-push-lib.js:384:31)\n' +
  '    at ClientRequest.emit (events.js:311:20)\n' +
  '    at ClientRequest.EventEmitter.emit (domain.js:482:12)\n' +
  '    at TLSSocket.emitRequestTimeout (_http_client.js:714:9)\n' +
  '    at Object.onceWrapper (events.js:417:28)\n' +
  '    at TLSSocket.emit (events.js:311:20)\n' +
  '    at TLSSocket.EventEmitter.emit (domain.js:482:12)\n' +
  '    at TLSSocket.Socket._onTimeout (net.js:478:8)\n' +
  '    at listOnTimeout (internal/timers.js:549:17)\n' +
  '    at processTimers (internal/timers.js:492:7)',
message: 'Socket timeout'
 },
```

carpasse:
@marco-c I thought it would be nice to be able to set a max time for the request while at the same time be able to drop the request earlier on connection drop. I can change the PR to only set the socketTimeout if you prefer

marco-c:
> @marco-c I thought it would be nice to be able to set a max time for the request while at the same time be able to drop the request earlier on connection drop. I can change the PR to only set the socketTimeout if you prefer

Yeah, I'd only keep `socketTimeout` (updating the description, but keeping it with the same name as before, `timeout`, for compatibility reasons). If we see a feature request for the overall timeout too, we can add it in the future (but I'd prefer not to add complexity unless it's a really useful feature).

carpasse:
Sure, I've updated the PR leaving socketTimeout logic but using the existing `timeout` option to set its value. Please let me know if you want other changes.

PR #612, opened 2020-08-04
Add run on repl.it badge to README

whyiwant:
This pull request adds a  badge to the . This will allow users to easily run this repository in their browser, without having to set up an environment. You can learn more about Repl.it [here](https://repl.it).

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "nodejs"
+run = "npm install -g web-push"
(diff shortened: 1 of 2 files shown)


marco-c:
Unfortunately I think this library isn't great for repl.it, as you need a complex scenario (run a browser, register a service worker, etc., etc.).

PR #787, opened 2023-03-31
add documentation for setVapidDetails method

delfuego:
I built a sample project today for my colleagues that uses web-push, and noticed that the `setVapidDetails()` method isn't documented, so I added that.

In addition, during the building of my project, I noticed that everything worked fine until I got to Safari, which rejected attempts to send a new push notification to a subscriber endpoint with a `BadJwtToken` error. Debugging it, I found that Apple's push notification endpoint used by Safari (at least Safari v16.4 on macOS) appears to not allow `https://localhost` URIs as VAPID subjects; changing my VAPID subject to either a `mailto:` URI or any `https:` URI that wasn't `localhost` allowed Safari to succeed, but using `https://localhost` (in any way at all, e.g., with or without port numbers) reverted back to the push notification endpoint rejecting the request with `BadJwtToken`.

(Note that if you think it's worth parsing for this in the `validateSubject` method of `vapid-helper.js`, and emitting a warning to the log, I could add a PR for that as well.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,13 +84,13 @@ Then you can run the following commands:
   ```
    The subscription value:
   ```javascript
-    { 
-      "endpoint": "https://fcm.googleapis.com/fcm/send/d61c5u920dw:APA91bEmnw8utjDYCqSRplFMVCzQMg9e5XxpYajvh37mv2QIlISdasBFLbFca9ZZ4Uqcya0ck-SP84YJUEnWsVr3mwYfaDB7vGtsDQuEpfDdcIqOX_wrCRkBW2NDWRZ9qUz9hSgtI3sY", 
-      "expirationTime": null, 
-      "keys": { 
-        "p256dh": "BL7ELU24fJTAlH5Kyl8N6BDCac8u8li_U5PIwG963MOvdYs9s7LSzj8x_7v7RFdLZ9Eap50PiiyF5K0TDAis7t0", 
-        "auth": "juarI8x__VnHvsOgfeAPHg" 
-      } 
+    {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -84,13 +84,13 @@ Then you can run the following commands:
...
    The subscription value:
   ```javascript
-    { 
-      "endpoint": "https://fcm.googleapis.com/fcm/send/d61c5u920dw:APA91bEmnw8utjDYCqSRplFMVCzQMg9e5XxpYajvh37mv2QIlISdasBFLbFca9ZZ4Uqcya0ck-SP84YJUEnWsVr3mwYfaDB7vGtsDQuEpfDdcIqOX_wrCRkBW2NDWRZ9qUz9hSgtI3sY", 
-      "expirationTime": null, 
-      "keys": { 
-        "p256dh": "BL7ELU24fJTAlH5Kyl8N6BDCac8u8li_U5PIwG963MOvdYs9s7LSzj8x_7v7RFdLZ9Eap50PiiyF5K0TDAis7t0", 
-        "auth": "juarI8x__VnHvsOgfeAPHg" 
-      } 
+    {
```
marco-c, line 87: Nit: please discard the whitespace changes from the patch, so we have a minimal diff and cleaner history

delfuego: Eek — I didn't notice those. Sorry. :(

delfuego:
Ok, I’ll add that to this PR later today, and clean up the white space accident, and resubmit!

delfuego:
OK, white space is cleaned up, and there's now a check for localhost https URIs as the VAPID subject in the validator. Notes:

* the check does not throw an error if a localhost https URI is found, but rather, just emits a console warning (since it's accepted by all the major browser's push notification servers but Apple's);
* the check specifically only emits a warning with **https** localhost URIs; in my testing, Apple's server is perfectly happy accepting **mailto** localhost URIs (e.g., `mailto:example@localhost`).

I thought about also including a check in the `validateSubject()` method for whether the subject is a URI that's not https: or mailto:, since the VAPID spec explicitly only allows those two URI types, but that's out of scope for this PR so I didn't include it.

marco-c:
> I thought about also including a check in the `validateSubject()` method for whether the subject is a URI that's not https: or mailto:, since the VAPID spec explicitly only allows those two URI types, but that's out of scope for this PR so I didn't include it.

Makes sense, I'd be happy to accept that in another PR ;)

marco-c (approved):

PR #789, opened 2023-03-31
tighten Vapid subject validation

delfuego:
only accept Vapid subjects with https: or mailto: URL protocols, as per spec

Note that I moved to using the WHATWG URL parser, which has been supported in Node since v6 (the min version supported as per `package.json`) since Node says that `url.parse()` is now fully-deprecated. I did local testing all the way down to Node 6, but the built-in unit tests can only run in Node 12+ due to various dev dependencies of the project that either aren't supported or won't run below that version; in Node 12+, the built-in unit tests all pass (and I added two additional ones for invalid, non-https:/non-mailto: URLs).


test/test-vapid-helper.js
```diff
@@ -10,7 +10,9 @@ const vapidHelper = require('../src/vapid-helper');
 
 const VALID_AUDIENCE = 'https://example.com';
 const VALID_SUBJECT_MAILTO = 'mailto: example@example.com';
-const VALID_SUBJECT_URL = 'https://exampe.com/contact';
+const VALID_SUBJECT_URL = 'https://example.com/contact';
+const INVALID_SUBJECT_URL_1 = 'http://example.gov';
+const INVALID_SUBJECT_URL_2 = 'ftp://example.net';
```
marco-c, line 15: As a test, could you also add localhost mailto and https urls?

src/vapid-helper.js
```diff
@@ -71,18 +71,22 @@ function validateSubject(subject) {
...
-      throw new Error('Vapid subject is not a url or mailto url. ' + subject);
-    } else if (subjectParseResult.hostname === 'localhost' && subjectParseResult.protocol === 'https:') {
-      console.warn('VAPID subject points to a localhost web URI, which is unsupported by Apple\'s push notification '
-      + 'server and will result in a BadJwtToken error when sending notifications.');
+  try {
+    const subjectParseResult = new URL(subject);
+    if (!['https:', 'mailto:'].includes(subjectParseResult.protocol)) {
+      throw new Error('Vapid subject is not an https: or mailto: URL. ' + subject);
     }
+    if (subjectParseResult.hostname === 'localhost') {
```
marco-c, line 83: Doesn't this also warn in case of mailto?

test/test-vapid-helper.js
delfuego: Sure — but just an FYI, they both should pass, since localhost mailto: URIs are accepted by Apple's push notification server, and localhost https: URIs just generate a console warning but no thrown error, and I don't believe we can test for console output without a bunch of changes to your Mocha test harness.

src/vapid-helper.js
delfuego: No; the WHATWG URL parser doesn't put the mailto: hostname into the hostname property, it ends up in the `pathname` property (by design). See [here](https://runkit.com/delfuego/url-parsing-next-gen).

marco-c (approved):
Thanks!

PR #786, opened 2023-03-30
migrate urlsafe-base64 to be internal to web-push

hikarunoryoma:
Solves: https://github.com/web-push-libs/web-push/issues/785

PR #812, opened 2023-06-10
Fix for the #786

haouarihk:
FIxed the issue with decode #786


haouarihk:
i hate macos 16 for this

haouarihk:
i just corrected a little mistake @hikarunoryoma made with 
`Buffer.from` he did `buffer.from(base64).toString("base64")` which essentially takes in the base64 as a utf-8 and tries to convert it into a string of type "base64"

but instead it should be
`buffer.from(base64, "base64).toString()` which is telling the buffer object to take that base64 and tell it it's a type of base64, and then stringify that buffer

.gitignore
```diff
@@ -1,3 +1,9 @@
 coverage/
 node_modules/
 test/output/
+
+# ignore nyc output
+.nyc_output
+
+# ignore history
+.history
```
marco-c, line 9: These changes in .gitignore are OK, but can you create a separate PR for them? I prefer to have self-contained PRs.

src/vapid-helper.js
```diff
@@ -100,14 +100,14 @@ function validatePublicKey(publicKey) {
...
+  if (!urlBase64Helper.validate(publicKey)) {
     throw new Error('Vapid public key must be a URL safe Base 64 (without "=")');
   }
 
-  publicKey = urlBase64.decode(publicKey);
+  publicKey = urlBase64Helper.decode(publicKey);
 
   if (publicKey.length !== 65) {
-    throw new Error('Vapid public key should be 65 bytes long when decoded.');
+    throw new Error(`Vapid public key should be 65 bytes long when decoded. got (${publicKey.length}) ${publicKey}`);
```
marco-c, line 110:
```suggestion
    throw new Error(`Vapid public key should be 65 bytes long when decoded, got (${publicKey.length}) ${publicKey}`);
```

.gitignore
haouarihk: i see sorry

marco-c: No problem, thank you for working on the PR!

BTW, you don't need to close and create a new one, you can also just update an existing PR.

hikarunoryoma:
Thanks for addressing here!

PR #813, opened 2023-06-11
Base64 decode fix

haouarihk:
FIxed the issue with decode https://github.com/web-push-libs/web-push/pull/786

# Explained
I just corrected a little mistake @hikarunoryoma have made with
`Buffer.from` he did `buffer.from(base64).toString("base64")` which essentially takes in the `base64` as a utf-8 and tries to convert it into a string of type "base64"

but instead it should be
`buffer.from(base64, "base64")` which is telling the buffer object to take that `base64` and tell it it's a type of base64


src/urlsafe-base64-helper.js
```diff
@@ -0,0 +1,32 @@
...
+function decode(base64) {
+   // Add removed at end '='
+  base64 += Array(5 - (base64.length % 4)).join('=');
+
+  base64 = base64
+    .replace(/-/g, '+') // Convert '-' to '+'
+    .replace(/_/g, '/'); // Convert '_' to '/'
+
+  // change from urlsafe-base64 since new Buffer() is deprecated
+  return Buffer.from(base64, "base64").toString();
```
marco-c, line 21: Need to use single quotes here

haouarihk: fixed that

marco-c:
There are some test failures, strangely they were not present in your earlier PR. Maybe you changed something?

src/urlsafe-base64-helper.js
```diff
@@ -0,0 +1,32 @@
...
+function decode(base64) {
+   // Add removed at end '='
+  base64 += Array(5 - (base64.length % 4)).join('=');
+
+  base64 = base64
+    .replace(/-/g, '+') // Convert '-' to '+'
+    .replace(/_/g, '/'); // Convert '_' to '/'
+
+  // change from urlsafe-base64 since new Buffer() is deprecated
+  return Buffer.from(base64, 'base64').toString();
```
marco-c, line 21: In your earlier PR, you were doing `return Buffer.from(base64, 'base64');`: https://github.com/web-push-libs/web-push/pull/812/files#diff-86818fdabcd5c8f177c6c670df721fd6c2659a9d736388ca36e24a6482ad9433R21.

haouarihk: Aah.
I'm clumsy
I will fix it right away

marco-c (approved):
Thanks!

PR #819, opened 2023-06-19
Use builtin encoding `base64url`

nikeee:
Node 16+ supports base64url:
https://nodejs.org/docs/latest-v16.x/api/buffer.html


src/urlsafe-base64-helper.js
```diff
@@ -1,32 +1,31 @@
...
-  base64 += Array(5 - (base64.length % 4)).join('=');
-
-  base64 = base64
-    .replace(/-/g, '+') // Convert '-' to '+'
-    .replace(/_/g, '/'); // Convert '_' to '/'
-
-  // change from urlsafe-base64 since new Buffer() is deprecated
-  return Buffer.from(base64, 'base64');
+  return Buffer.from(base64, 'base64url');
 }
```
marco-c, line 17: Could you just remove these functions now and use `buffer.toString` and `Buffer.from` directly in the call sites?

nikeee: Done. I left `validate` in there, because there is no equivalent for that.

.gitignore
```diff
@@ -1,3 +1,5 @@
 coverage/
 node_modules/
 test/output/
+.nyc_output/
```
marco-c, line 4: This is an acceptable change, but could you do it in a separate PR?
I prefer to have each PR be self-contained

nikeee: Done.
Maybe we should squash this when merging. Didn't have the time to do that.

marco-c (approved):
Thanks!

PR #828, opened 2023-07-22
fix(sendnotification): fix HttpsProxyAgent error

neilning-xc:
https-proxy-agent 7.x changed the default export
https://github.com/TooTallNate/proxy-agents/blob/main/packages/https-proxy-agent/README.md#https-module-example

### DIFF
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -362,7 +362,7 @@ WebPushLib.prototype.sendNotification = function(subscription, payload, options)
       }
 
       if (requestDetails.proxy) {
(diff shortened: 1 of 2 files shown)


marco-c:
The fix looks good to me, but the test is using a proxy that doesn't exist. If we want to test, we actually need to run a proxy locally as part of the tests.

neilning-xc:
> The fix looks good to me, but the test is using a proxy that doesn't exist. If we want to test, we actually need to run a proxy locally as part of the tests.

Oh, I forgot that. I remove that test case if you don't mind

marco-c (approved):

marco-c:
Thanks!

PR #841, opened 2023-08-28
Fixes sendNotification named import declaration

### LINKED ISSUE #683: Cannot read property 'generateRequestDetails' of undefined on TypeScript version greater than 4.0.
Got problem with web-push package using with typescript version greater than 4.0 but it works fine on typescript versions lesser than 4.0.
Environment info:
-   NodeJs Version: 14.17.6
-   Yarn Version: 1.22.11
-   web-push package version: 3.4.5

tsconfig.json:

> {
    "compilerOptions": {
        "outDir": "../out",
        "module": "commonjs",
        "target": "esnext",
        "moduleResolution": "node",
        "emitDecoratorMetadata": true,
        "experimentalDecorators": true,
        "sourceMap": true,
        "skipLibCheck": true,
        "declaration": false,
        "resolveJsonModule": true,
        "lib": [
            "dom",
            "esnext",
            "esnext.asynciterable"
        ]
    },
    "include": [
        "./**/*"
    ]
}

Error Message:
`TypeError: Cannot read property 'generateRequestDetails' of undefined`

brajkowski:
Fixes #683

I was able to reproduce the TypeScript issue seen in #683 within the existing test suite in 30c5932.  Previously, requiring `sendNotification` as a named import would fail [here](https://github.com/web-push-libs/web-push/blob/master/src/web-push-lib.js#L341) due to the function lacking reference to the prototype instance.  5277eea binds the prototype instance to the exported `sendNotification` function using [ES5 syntax](https://caniuse.com/es5).  The pattern of testing on named imports is extended to all other `WebPushLib` methods in 2548c3a to prevent the same issue from occurring on another method in the future.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,5 +17,5 @@ module.exports = {
   setGCMAPIKey: webPush.setGCMAPIKey,
   setVapidDetails: webPush.setVapidDetails,
   generateRequestDetails: webPush.generateRequestDetails,
-  sendNotification: webPush.sendNotification
+  sendNotification: webPush.sendNotification.bind(webPush)
 };
--- a/test/test-generate-request-details.js
+++ b/test/test-generate-request-details.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const assert = require('assert');
-const webPush = require('../src/index');
+const { generateRequestDetails } = require('../src/index');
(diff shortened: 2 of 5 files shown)


marco-c (approved):
Thanks!

PR #843, opened 2023-08-30
Fix validation error of protocol not beeing returned

### LINKED ISSUE #830: Validation error of protocol is never returned
The validation error at line 81 is never returned, it's overwritten by the error in the catch at line 89.

https://github.com/web-push-libs/web-push/blob/42179be3928d73879a6a06e7b6a9b0be9b1b7d53/src/vapid-helper.js#L78-L90

glasgio:
Fixes #830

Fix the "https: or mailto:" error being overwritten by the error in the catch by passing it upwards trough the cause property

### DIFF
--- a/src/vapid-helper.js
+++ b/src/vapid-helper.js
@@ -75,8 +75,9 @@ function validateSubject(subject) {
     + 'mailto: address. ' + subject);
   }
 
+  let subjectParseResult = null;
   try {
-    const subjectParseResult = new URL(subject);
+    subjectParseResult = new URL(subject);
   } catch (err) {
     throw new Error('Vapid subject is not a valid URL. ' + subject);
   }
--- a/src/vapid-helper.js
+++ b/src/vapid-helper.js
@@ -77,17 +77,17 @@ function validateSubject(subject) {
 
   try {
     const subjectParseResult = new URL(subject);
-    if (!['https:', 'mailto:'].includes(subjectParseResult.protocol)) {
-      throw new Error('Vapid subject is not an https: or mailto: URL. ' + subject);
-    }
-    if (subjectParseResult.hostname === 'localhost') {
-      console.warn('Vapid subject points to a localhost web URI, which is unsupported by '
-        + 'Apple\'s push notification server and will result in a BadJwtToken error when '
-        + 'sending notifications.');
     }
   } catch (err) {
-    throw new Error('Vapid subject is not a valid URL. ' + subject, { cause: err });
+    throw new Error('Vapid subject is not a valid URL. ' + subject);
   }
+  if (!['https:', 'mailto:'].includes(subjectParseResult.protocol)) {
+    throw new Error('Vapid subject is not an https: or mailto: URL. ' + subject);
+  }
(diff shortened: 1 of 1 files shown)


src/vapid-helper.js
```diff
@@ -86,7 +86,7 @@ function validateSubject(subject) {
         + 'sending notifications.');
     }
   } catch (err) {
-    throw new Error('Vapid subject is not a valid URL. ' + subject);
+    throw new Error('Vapid subject is not a valid URL. ' + subject, { cause: err });
   }
```
marco-c, line 90: What if we moved the try-catch to only wrap the `const subjectParseResult = new URL(subject);` line?

glasgio: I think it would work as well since it probably was the main reason the try-catch was put there in the first place. And it will also make the error on the console more clean/targetable.

I added the cause solution trying to change as minimun as possible

marco-c: I'd prefer moving the try-catch, it means a bit more changes, but it seems nicer to avoid throwing an exception only to catch it and rethrow another exception with a different message.

src/vapid-helper.js
```diff
@@ -77,17 +77,17 @@ function validateSubject(subject) {
...
   try {
     const subjectParseResult = new URL(subject);
-    if (!['https:', 'mailto:'].includes(subjectParseResult.protocol)) {
-      throw new Error('Vapid subject is not an https: or mailto: URL. ' + subject);
-    }
-    if (subjectParseResult.hostname === 'localhost') {
-      console.warn('Vapid subject points to a localhost web URI, which is unsupported by '
-        + 'Apple\'s push notification server and will result in a BadJwtToken error when '
-        + 'sending notifications.');
     }
```
marco-c, line 80: there's an additional `}` here that causes a syntax error

marco-c (approved):

marco-c:
Thanks!

marco-c:
Oops, there is still some error:
> /home/runner/work/web-push/web-push/src/vapid-helper.js
> Error:   79:11  error  'subjectParseResult' is assigned a value but never used  no-unused-vars
> Error:   83:39  error  'subjectParseResult' is not defined                      no-undef
> Error:   86:7   error  'subjectParseResult' is not defined                      no-undef

glasgio:
> Oops, there is still some error:
> 
> > /home/runner/work/web-push/web-push/src/vapid-helper.js
> > Error:   79:11  error  'subjectParseResult' is assigned a value but never used  no-unused-vars
> > Error:   83:39  error  'subjectParseResult' is not defined                      no-undef
> > Error:   86:7   error  'subjectParseResult' is not defined                      no-undef

It makes sense since the subjectParseResult is now scoped, sorry. Do you prefer an approach  like defining the variable before the try?

`
let subjectParseResult = undefined
  try {
    subjectParseResult = new URL(subject);
  } catch (err) {
    throw new Error('Vapid subject is not a valid URL. ' + subject);
  }`

marco-c (approved):

PR #844, opened 2023-09-02
adds .nyc_output/ to .gitignore

brajkowski:
`npm test` generates a `.nyc_output` directory that shouldn't be committed to the repository.


marco-c (approved):
Thanks!

PR #849, opened 2023-09-14
Remove link to serviceworke.rs

emhoracek:
According to https://github.com/mdn/serviceworker-cookbook/pull/335,  "The domain https://serviceworke.rs/ (do not open) has been overtaken and are used to steal clicks. This removes the domain and uses a dummy domain."


marco-c (approved):
Thanks!

Issue #772, opened 2023-02-08
support ESM


eng-dibo:
I want to use `web-push` server functions i.e. `generateVAPIDKeys` and `sendNotification`, but it seems that web-push doesn't support esm packages, i.e packages that defines `type: module` on its package.json

```
SyntaxError: Named export 'generateVAPIDKeys' not found. The requested module 'web-push' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
```

when using the default export the same error occurs because the file is compiled into ESM

MumiaIrrequieta:
Try this @eng-dibo :
`import webpush from 'web-push'

webpush.sendNotification(...)`

senbar:
I migrated everything to ESM but have one problem with tests namely testSendNotification.js had before each 
```js
 mocha.beforeEach(function () {
    ({ sendNotification, setGCMAPIKey, setVapidDetails } = require('../src/index'));
  });```
and then it deleted cache before each for clean start
```js
  mocha.beforeEach(function() {
    requestBody = null;
    requestDetails = null;

    // Delete caches of web push libs to start clean between test runs
    delete require.cache[path.join(__dirname, '..', 'src', 'index.js')];
    delete require.cache[path.join(__dirname, '..', 'src', 'web-push-lib.js')];
```
How would you recommend me approaching this if we wnat to get rid of require to make it ESM?

marco-c:
I'm not sure to be honest, I'm not familiar enough with ESM.

senbar:
Would you be ok with me adding clearCache method to web-push-lib.js and index.js that would clear all state for testing? I don't think there's other way around that

marco-c:
What if we simply call `setGCMAPIKey` and `setVapidDetails` with `null` as a way to reset the state?

nikeee:
Maybe if #876 is done first, the need to reset state vanishes.

khanyasar022:
> I migrated everything to ESM but have one problem with tests namely testSendNotification.js had before each
> 
> ```js
>  mocha.beforeEach(function () {
>     ({ sendNotification, setGCMAPIKey, setVapidDetails } = require('../src/index'));
>   });
> ```
> 
> and then it deleted cache before each for clean start
> 
> ```js
>   mocha.beforeEach(function() {
>     requestBody = null;
>     requestDetails = null;
> 
>     // Delete caches of web push libs to start clean between test runs
[...]

PR #877, opened 2024-01-08
Changed from commonJS module to ESM.

senbar:
Related #772

Changed commonJS modules to ESM. Someone please double run this since I had to run this is docker container (selenium assistant doesn't support windows and that's the only thing I have available rn) and only headless tests run.

The problem is clearing imported modules in beforeEach test I mentioned in #772 , how should I solve this?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,9 @@
   "env": {
     "es6": true
   },
+  "parserOptions": {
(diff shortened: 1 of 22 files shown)


pierredarrieutort (approved):

marco-c:
This is being handled in #961.

PR #889, opened 2024-02-20
Support multiple instances

### LINKED ISSUE #876: Add support for multiple WebPush instances
This is a feature requests for a class-based update of this amazing library.

For example:

```js
import { WebPush } from 'web-push';

const webPush = new WebPush();
webpush.setVapidDetails(...);
webpush.sendNotification(...);
```

This will allow for using multiple credentials in a single Node.js server. It also feels like cleaner code in my opinion.

lokshunhung:
## Goal
Resolves #876

## Implementation
- Instead of having a module-scoped `gcmAPIKey` and `vapidDetails` in `src/web-push-lib.js`, we change those to instance properties of a `WebPushLib` class instance.
- When re-exporting the `setGCMAPIKey`, `setVapidDetails`, `generateRequestDetails` for the global instance of `WebPushLib`, we need to `.bind()` the methods to retain the `this` context.
- Also export the `WebPushLib` constructor so users can create multiple instances instead of just using the global instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,32 @@ object will contain:
 
 <hr />
 
+# Using multiple instances of `WebPushLib`
+
+The functions `setGCMAPIKey`, `setVapidDetails`, `generateRequestDetails`, `sendNotification`
+exported by this module are bounded to a global instance of `WebPushLib`.
+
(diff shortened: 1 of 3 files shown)


marco-c:
@lokshunhung could you add a test for this?

Issue #943, opened 2026-01-02
Deprecation warning of url.parse in Node 24


GabrielGil:
Hello, I am getting some deprecation errors:

```shell
(node:4) [DEP0169] DeprecationWarning: url.parse() behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for url.parse() vulnerabilities.
```

These seem to appear when sending notifications, at least, from https://github.com/web-push-libs/web-push/blob/master/src/web-push-lib.js.

I've patched my version locally with pnpm, in case it comes handy for someone else:

`// File: patches/web-push@3.6.7.patch`
```
diff --git a/src/web-push-lib.js b/src/web-push-lib.js
index a47df054e7d9809cde27d6e645d66f60b84dd8c1..b27f70c9c2d2a5456852825a4d8c71b612c587ba 100644
--- a/src/web-push-lib.js
+++ b/src/web-push-lib.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const url = require('url');
+// const url = require('url'); // Removed - using WHATWG URL API instead
 const https = require('https');
 
 const WebPushError = require('./web-push-error.js');
@@ -271,7 +271,7 @@ WebPushLib.prototype.generateRequestDetails = function(subscription, payload, op
         requestDetails.headers.Authorization = 'key=' + currentGCMAPIKey;
       }
     } else if (currentVapidDetails) {
-      const parsedUrl = url.parse(subscription.endpoint);
+      const parsedUrl = new URL(subscription.endpoint);
       const audience = parsedUrl.protocol + '//'
       + parsedUrl.host;
 
@@ -345,10 +345,10 @@ WebPushLib.prototype.sendNotification = function(subscription, payload, options)
 
     return new Promise(function(resolve, reject) {
       const httpsOptions = {};
-      const urlParts = url.parse(requestDetails.endpoint);
+      const urlParts = new URL(requestDetails.endpoint);
       httpsOptions.hostname = urlParts.hostname;
-      httpsOptions.port = urlParts.port;
-      httpsOptions.path = urlParts.path;
+      httpsOptions.port = urlParts.port || undefined;
+      httpsOptions.path = urlParts.pathname + urlParts.search;
 
       httpsOptions.headers = requestDetails.headers;
       httpsOptions.method = requestDetails.method;

```

but it'd be great to update the lib, not sure if this is in the plans.

Using version 3.6.7.

marco-c:
[...]

PR #945, opened 2026-04-11
fix: replace deprecated url.parse() with new URL()

mriganshu16:
Node.js 24 throws a deprecation warning (DEP0169) when url.parse() 
is used. This replaces the two usages of url.parse() in web-push-lib.js 
with the modern WHATWG URL API (new URL()), which is the recommended 
replacement.

Fixes the issue reported in #943


src/web-push-lib.js
janvorwerk, line 3: Is this still required?

RetfalviBence: Hi @mriganshu16, can you update the PR?

alexneventchin:
Bump, this should get merged

marco-c:
This was done in #948.

PR #948, opened 2026-06-17
chore: bump jws and test certificate

### LINKED ISSUE #943: Deprecation warning of url.parse in Node 24

43081j:
Does two things:

1. bump `jws` to support modern Node
2. create a new test cert which complies with newer openssl constraints (key size)

This is part of me trying to modernise the repo and what not.

**I have many stacked PRs ready if this one lands, and some replace currently open PRs.**


package.json
```diff
@@ -42,9 +42,9 @@
     "eslint-config-airbnb": "19.0.4",
     "eslint-plugin-import": "2.29.1",
     "geckodriver": "4.3.0",
-    "nyc": "15.1.0",
     "mkdirp": "3.0.1",
     "mocha": "10.4.0",
+    "nyc": "15.1.0",
```
marco-c: nit: could you move this back?

marco-c:
@43081j thanks, lgtm. In other PRs, we'll need to fix the CI.

marco-c:
@43081j ideally, I'd like to have separate PRs or at least commits for self-contained changes (e.g. the URL thing and the jws thing).

43081j:
> @43081j ideally, I'd like to have separate PRs or at least commits for self-contained changes (e.g. the URL thing and the jws thing).

What you're describing is exactly how every branch I have is. I think this is just a rare occurrence, I probably thought the URL change was needed at the time, I don't remember 👍

I will submit the rest next time I'm at a laptop.

package.json
43081j: This wasn't me, npm orders them that way when the dependencies change. 

We can change it back but I feel like we shouldn't fight the preferred order of npm

43081j:
i've caught up from main but need you to approve CI again. it does pass locally so im just going to have to wait for CI results 👀 

also FYI @marco-c i would hold off merging old PRs that might touch dependencies, modernisation etc. i have 7 branches including this one - most things are probably accounted for.

i'll open the rest knowing some are stacked, for visibility at least

package.json
marco-c: I agree, but I'd like to do that in a separate PR!

PR #949, opened 2026-07-10
test: switch tests to use playwright

43081j:
Depends on #948

**This is a stacked PR**. The diff won't be of much use until #948 merges. The individual commits can be reviewed though.

This switches from selenium/webdriver to playwright for tests. 

The test setup has changed a lot because of that, but the individual test cases are pretty much the same as before.

**Important:**

- Tests fail on master in CI afaict
- Firefox is skipped here because headless firefox (that puppeteer, playwright, etc pull down) doesn't support web push (it has no keys in it for the push service)


marco-c:
@43081j could you resolve conflicts?

marco-c:
> Firefox is skipped here because headless firefox (that puppeteer, playwright, etc pull down) doesn't support web push (it has no keys in it for the push service)

That's sad, we need to figure out if we can fix that.

43081j:
Locally you can run it against a normal install of Firefox, but none of the packaged ones support it. 

The browser tests are a bit awkward since they're e2e. They're also testing that each browser's push service is working, really. That feels like something which will forever be flaky since we're depending on a third party service during tests. 

I do wonder if we should just mock it out instead. So then we just test that our code calls the browser API and assume the rest works.

marco-c:
They were meant to be integration tests, to make sure browsers implementation doesn't change in a way that breaks the library.
They have rarely, if ever, failed due to web push services being unavailable.

43081j:
that's fair. they would have caught the changes in required key type for example.

there's not a lot we can do about the firefox ones other than install regular firefox through `apt` or something and point the tests at that.

i think we should land this one then investigate that as a follow up (since we'd need this setup anyway)

marco-c:
Yes, agreed. Could you file a follow-up to figure out how to make tests work in Firefox?

test/testSelenium.js
```diff
@@ -1,292 +0,0 @@
-'use strict';
```
marco-c, line 1: git is not great at keeping history here if we rename the file and make changes at the same time. Could you rename the file in a commit, and make the modifications in another?
This will also help me review better, in addition of keeping a better git blame.

43081j: done but doesn't make much of a difference to the overall diff. though you can at least review by commit now

marco-c (approved):

marco-c:
@43081j I've merged this, but tests are still failing for now.

PR #961, opened 2026-07-14
feat: migrate package to ESM

43081j:
Depends on #949 - this PR is stacked so better off reviewing it once that lands, or by commit

This changes the `type` to be `"module"` and rewrites the sources to be ES modules.

All LTS of node (and some previous versions, since 20.x) support `require(esm)` these days. So both CJS and ESM users can import ESM just fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "web-push",
   "version": "3.6.7",
   "description": "Web Push library for Node.js",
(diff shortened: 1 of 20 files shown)


marco-c:
@43081j could you rebase?

marco-c (approved):

PR #965, opened 2026-07-21
test: run tests using xvfb

43081j:
In order for Linux tests to pass, we need to run chrome with `xvfb-run`
to keep it headed.

Eventually it'd be nice to go headless but it seems Chrome doesn't
enable push in headless mode. `headless: 'new'` may work for that
eventually.


43081j:
@marco-c this should fix CI test runs.

though coveralls is falling over for whatever reason now on macOS.

marco-c (approved):

marco-c:
Let's file a follow-up for the mac coveralls problem.

PR #972, opened 2026-07-24
feat: add typescript types

43081j:
Adds the correct JSDoc so we can use typescript to generate `d.ts` from it.

@marco-c this is more a nice-to-have. we have good enough JSDoc to generate the types so it might be nice to do so and publish them.

if you don't think we should, though, let me know.

### DIFF
--- a/src/asn1.d.ts
+++ b/src/asn1.d.ts
@@ -1,0 +1,57 @@
+declare module 'asn1.js' {
+  interface Asn1Node {
+    seq(): Asn1Node;
+    obj(...children: Asn1Node[]): Asn1Node;
+    key(name: string): Asn1Node;
+    explicit(tag: number): Asn1Node;
+    optional(): Asn1Node;
+    int(): Asn1Node;
+    octstr(): Asn1Node;
+    objid(): Asn1Node;
+    bitstr(): Asn1Node;
+  }
+
+  interface DecodeOptions {
+    partial?: boolean;
+    track?: (path: string, start: number, end: number, type: string) => void;
+  }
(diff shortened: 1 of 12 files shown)


src/http_ece.d.ts
```diff
@@ -0,0 +1,33 @@
+declare module 'http_ece' {
```
marco-c, line 1: Can we add these in http_ece?

src/asn1.d.ts
```diff
@@ -0,0 +1,57 @@
+declare module 'asn1.js' {
```
marco-c, line 1: Can we add these in asn1.js?

src/http_ece.d.ts
43081j: its in the web-push-libs org so presumably we can land a PR there. i can open one when i get chance 👍 

it doesn't need to block this for now though IMO

src/asn1.d.ts
43081j: it hasn't changed in many years so i'm not very confident we will manage to add them. i can open an issue though

marco-c: Let's try, you never know!

src/http_ece.d.ts
marco-c: I wouldn't block on the asn1 one, but maybe on this one yes, so we don't need to do another PR to remove them once it's done in http_ece directly. WDYT?

43081j: Sure I can try find time to look 👍

PR #991, opened 2026-09-18
docs: explain how to handle sendNotification() error status codes

SrijaVuppula:
`sendNotification()` currently documents that a rejection carries `statusCode`,
`headers`, and `body`, but doesn't explain what to do with them. This adds a
short "Handling errors" section covering the status codes that matter most in
practice — distinguishing a permanently dead subscription (404/410, safe to
delete) from a transient failure worth retrying (429, 5xx) — with a small
code example.

No code changes, README only.


marco-c:
Can you reference the web push standards for the 4** errors?

